PR #15, opened 2018-04-19
add a travis badge

thraxil:
Everybody loves badges.


thraxil:
Ah. nevermind. I see there is already one there.

PR #21, opened 2018-05-04
Feature/api hookup

grozdanowski:
This PR adds the following:

- the Dashboard is now located at the root URL (as it should be)
- certain small style fixes (clashes with the Open edX stylesheet)
- adds actions and reducers to Redux so it can store/fetch lists of courses and users
- populate the said Redux state by calling the API
- display the course list retrieved from the API in the course selector dropdown
- display actual course data in the single course page header area


johnbaldwin:
Merging into `john/course-enrollment` so we can test/dev integration of UI and REST API

PR #29, opened 2018-06-19
Added General User Data - Have baseline 'general user data' endpoint working

johnbaldwin:
- returns all users with the courses for which each user is enrolled
- filterable on the course id via query parameter
- Added tests, not 100% field coverage, but have tests for the view and
serializer

Ref: API doc https://docs.google.com/document/d/16_4R0Q9YQ7IbofjWTXLsdMK_v-loBC7CUoARXw5LRqo/

Which is really just this:
```
General User Data

[
  {
    "username": "maxi",
    "country": "UY",
    "is_active": true,
    "year_of_birth": 1985,
    "level_of_education": "b",
    "gender": "m",
    "date_joined": "2018-05-06T14:01:58Z",
    "language_proficiencies": [],
    "courses": [
      {
        "course_name": "Something",
        "course_id": "A193+2016Q4+something",
      }
      ...
    ]
  },
  ...
]

```

The implementation differs slightly in the following ways:

* The date joined is a date and not a datetime. Let me know if the time of day is important and I'll change it to be datetime. I find that excessive datetimes muddy it up when a date is sufficient

* The actual data returned for each course item contains the keys 'id', 'display_name', 'org', 'number'
This is because I'm reusing the CourseIndexSerializer
* added 'fullname' to the data 

# Testing

Test coverage can use improvement. I have a baseline, but didn't want to sink time into either cobbling together some crappy code coverage and didn't want to invest time (at this time) into doing "proper and robust" coverage for how users and courses (CourseOverview) are associated through non-direct model relationships between User, CourseEnrollment, and CourseOverview)

# Issues

The 'number' for the course is not being returned. I need to investigate and fix that. This is also the value 'course_code' specified in the general course data endpoint

### DIFF
--- a/devsite/requirements.txt
+++ b/devsite/requirements.txt
@@ -12,6 +12,9 @@
 python-dateutil==2.1
 path.py==8.2.1
 
+# Yes, this is old but is the one specified by Ginkgo edx-platform
+pytz==2016.7
+
 ##
 ## Django package dependencies
 ##
--- a/figures/filters.py
+++ b/figures/filters.py
@@ -129,3 +129,29 @@ class SiteDailyMetricsFilter(django_filters.FilterSet):
     class Meta:
         model = SiteDailyMetrics
         fields = ['date_for', 'date']
+
+
+class GeneralUserDataFilter(django_filters.FilterSet):
+
+    course_id = django_filters.MethodFilter(action='filter_course_id')
+
+    class Meta:
+        model = get_user_model()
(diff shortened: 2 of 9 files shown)


johnbaldwin:
[(5) Add "General User Data" REST API endpoint for frontend](https://trello.com/c/W0ppHhQj/434-5-add-general-user-data-rest-api-endpoint-for-frontend)

PR #34, opened 2018-08-28
Updated metrics to include missing history information

johnbaldwin:
- Added metrics methods to retrieve course specific metric data for a
date range
- Added convenience function to reuse the existing month metric data
retrieval function (We may refactor this to be an iterator)
- Updated REST API and front end to replace 'current' with
'current_month'
- Changed the front end 'averageCompletionTime' to be a number instead
of string

### DIFF
--- a/figures/metrics.py
+++ b/figures/metrics.py
@@ -24,13 +24,14 @@ parameter to support multi-tenancy
 from collections import namedtuple
 
 import datetime
+import math
 
 from dateutil.parser import parse as dateutil_parse
 from dateutil.rrule import rrule
 from dateutil.relativedelta import relativedelta
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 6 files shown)


johnbaldwin:
@grozdanowski If you would please review the changes I made to the UI code. I've tested this in my devstack

johnbaldwin:
[(3) API historic data missing current period](https://trello.com/c/ClCfY6z1/480-3-api-historic-data-missing-current-period)

johnbaldwin:
[(2) API - numeric values shouldn't return "null" if non existent data, but rather a 0 (zero)](https://trello.com/c/XXGbPvpN/481-2-api-numeric-values-shouldnt-return-null-if-non-existent-data-but-rather-a-0-zero)

PR #35, opened 2018-08-28
Merging develop to master

johnbaldwin:
We're still in pre-initial release, so if we need to fix something(s) then it won't break production (yet)


OmarIthawi:
@johnbaldwin I've removed myself from the reviewers since this is still a WIP. Please add me once this stops being a WIP.

bryanlandia:
@johnbaldwin Are you looking for reviews at this point since it's WIP? Could you check what failed with the Travis build?

johnbaldwin:
@bryanlandia What was failing where the Flake8  tests. Now it is just "failing" because codecov isn't at it's target, which I'm not super concerned about *right now* but I want to make sure that critical stuff is covered, so I have to do some local coverage checking to see if there's anything critical that I should add some last minute test coverage.


I *do* need to get this merged up to master this weekend, write install instructions, cut a release and post to PyPI so it is ready Monday (I also have to wrap up the envs mod and user menu addition in edx-platform so I can post that to PR

PR #46, opened 2018-10-08
Updated NPM package versions to address security issues

johnbaldwin:
Auto updated package.json and packag-lock.json with `npm audit fix`

NOTE: It is important to then delete `node_modules` and re-run as the
first call to `npm audit fix` may not catch all reported vulnerabilities


johnbaldwin:
Thanks!

PR #48, opened 2018-10-10
Updated Figures settings integration with the LMS + updated docs

johnbaldwin:
Updated how the settings integration works with Open edX LMS

* Figures now makes function calls to update settings
* Added a convenience function in figures.settings to wrap around the Webpack loader and CeleryBeat schedule settings update methods
* Added figures.settings.update_setings to figures/__init__ for convenience
* Added test for figures.settings
* Updated the devsite test settings so that it has the same Figures hooks as the LMS should have
* Updated install docs to match new settings integration with Open edX LMS

NOTE: I did NOT revamp the devstack install instructions to serve both developers and test drivers. I wanted to get this update out for review ASAP.

Also, I didn't include the version bump. I'll do that when I merge to develop and up to master

### DIFF
--- a/devsite/devsite/test_settings.py
+++ b/devsite/devsite/test_settings.py
@@ -78,5 +78,12 @@ ENV_TOKENS = {
 }
 
 
-# This is all we need to instantiate edx-figures with default settings
-from figures.settings import FIGURES
+# Enable Figures if it is included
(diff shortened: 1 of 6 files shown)


johnbaldwin:
Thanks @OmarIthawi !

PR #61, opened 2018-11-06
Updated devsite to run standalone

johnbaldwin:
This PR addresses decay in ignoring devsite after initial Figures development. We're doing this now because 

* We want devsite to run Figures demo mode 
* We want to be able to do some local development w/out requiring edx-platform

This is a simple baseline so that the built-in site, `devsite` will be able to serve Figures.

* Added registration and login so Figures can run unmodified with an
authenticated user
* Updated settings
* Updated templates


johnbaldwin:
[(2) Make devsite work (standalone mode)](https://trello.com/c/3JeBI9KI/649-2-make-devsite-work-standalone-mode)

devsite/devsite/settings.py
```diff
@@ -78,49 +82,70 @@
...
+        'rest_framework.renderers.BrowsableAPIRenderer',
+    )
+}
 
 
+# It expects them from the project's settings (django.conf.settings)
+WEBPACK_LOADER = {}
+CELERYBEAT_SCHEDULE = {}
+
+# Declare values we need from server vars (e.g. lms.env.json)
```
briandant: Does the empty dict mean that Figures does not necessarily require anything? Or is this documented elsewhere?

devsite/devsite/settings.py
```diff
@@ -78,49 +82,70 @@
...
+        'rest_framework.permissions.AllowAny',
+    ],
+    'DEFAULT_RENDERER_CLASSES': (
+        'rest_framework.renderers.JSONRenderer',
+        'rest_framework.renderers.BrowsableAPIRenderer',
+    )
+}
 
 
+# It expects them from the project's settings (django.conf.settings)
```
briandant: Which it? :)

devsite/devsite/settings.py
```diff
@@ -78,49 +82,70 @@
...
+WEBPACK_LOADER = {}
+CELERYBEAT_SCHEDULE = {}
+
+# Declare values we need from server vars (e.g. lms.env.json)
+ENV_TOKENS = {}
+
+# Enable Figures if it is included
+# This should be the same code as used in the LMS settings
```
briandant: So, the user has the option to let the code here do the work or to put it into LMS settings himself? In what case would the user want to do this? Could this be simplified by instructing the user to do it always or never?

devsite/devsite/settings.py
```diff
@@ -1,34 +1,30 @@
...
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 USE_TZ = True
+TIME_ZONE = 'UTC'
 ALLOWED_HOSTS = []
 
-
-# Application definition
+# Needed for Python to find the mock edx-platform modules
```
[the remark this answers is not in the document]
johnbaldwin: @briandant the edx-platform mocking is specific to Figures. Is this still a change you request?

devsite/devsite/settings.py
```diff
@@ -41,10 +37,18 @@
     'rest_framework',
     'django_countries',
     'django_filters',
-    #'rest_framework.authtoken',
+    # 'rest_framework.authtoken',
     'webpack_loader',
     'devsite',
     'figures',
+
+    # edx-platform apps. Mocks are used by default
```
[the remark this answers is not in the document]
johnbaldwin: Nope, don't have docs for testing yet. In the backlog. I can reword this (remove the 'used by default') as only mocks are used in devsite since its purpose is to provide a simple (and transparent to Figures) simulation of edx-platform

devsite/devsite/settings.py
johnbaldwin: To keep the figures include fragment identical to that used in real Open edX, I have the empty dict since `ENV_TOKENS` are part of the edx-platform settings.

No defaults are required to be set in `lms.env.json` as `figures.settings` defines the defaults. The `ENV_TOKENS` settings are for defaults overrides and likely in the future for declaring plugins.

Since you raised the question, I'll elaborate on the doc line here

devsite/devsite/settings.py
johnbaldwin: Figures, I'll clarify

devsite/devsite/settings.py
johnbaldwin: The code to load Figures in devsite is identical to the code specified in the installation instructions for the real edx-platform: https://github.com/appsembler/figures/blob/master/docs/source/install.rst

I can add that note since you asked the question

johnbaldwin:
@briandant I believe I addressed your comments

johnbaldwin:
@briandant Thanks for reviewing!

PR #65, opened 2018-12-18
Figures multisite support

johnbaldwin:
Production code + tests for review

# Context - Support both single site and multisite deployments

The code needs to support both single site and multisite deployments. Since single site deployments do not filter "per site" in queries, we can make 'Site' FKs mandatory for Figures metrics models and assign the default site for single site deployments. The other option considered is making the Site FK optional. 

# Up for review

*  Added settings handling, tests
* Added Site to metrics models
* Pipeline is now site aware. Populate sets of metrics per site. Initially will run all sites in sequence instead of as Celery subtasks. That will be a follow-on
* Added `sites` module as an facade to retrieve site filtered data
** Add awareness for custom `edx-organizations` (namely Appsembler's fork)
* Add multisite authentication and authorization to permissions module
* Add multisite support to views

The following items should not block getting Figures installed on Tahoe staging for testing, however, we *do* want to get this test coverage as we deploy to production

* Multisite test coverage for views
* Better testing for data leaking

### DIFF
--- a/figures/models.py
+++ b/figures/models.py
@@ -3,6 +3,7 @@
 """
 
 from django.conf import settings
+from django.contrib.sites.models import Site
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 
@@ -15,6 +16,10 @@ from model_utils.models import TimeStampedModel
 class CourseDailyMetrics(TimeStampedModel):
     """Metrics data specific to an individual course
 
+    CourseDailyMetrics instances are created before the SiteDailyMetrics. This,
+    along with the fact we now filter course metrics for a given site, we aren't
+    adding a SiteDailyMetrics foreign key. This is subject to change as the code
+    evolves.
     """
     date_for = models.DateField()
 
@@ -23,7 +28,7 @@ class CourseDailyMetrics(TimeStampedModel):
     # the CourseOverview model or have the course_id be a
     # CourseKeyField
     course_id = models.CharField(max_length=255)
-
+    site = models.ForeignKey(Site, default=settings.SITE_ID)
     enrollment_count = models.IntegerField()
     active_learners_today = models.IntegerField()
     # Do we want cumulative average progress for the month?
@@ -46,26 +51,36 @@ class CourseDailyMetrics(TimeStampedModel):
 @python_2_unicode_compatible
 class SiteDailyMetrics(TimeStampedModel):
     """
-    TODO: Add Multi-site support
-    add FK to site and make site + date_for unique together
+    Stores metrics for a given site and day
     """
 
     # Date for which this record's data are collected
-    date_for = models.DateField(unique=True)
+    date_for = models.DateField()
+    site = models.ForeignKey(Site, default=settings.SITE_ID)
     cumulative_active_user_count = models.IntegerField(blank=True, null=True)
     todays_active_user_count = models.IntegerField(blank=True, null=True)
     total_user_count = models.IntegerField()
     course_count = models.IntegerField()
     total_enrollment_count = models.IntegerField()
 
-    # Foreign key relationships
-    # site =
-
     class Meta:
-        ordering = ['-date_for']
+        """
+        SiteDailyMetrics view and serializer tests fail when we include 'site'
+        in the `unique_together` fields:
+
+            unique_together = ['site', 'date_for']
+
+            ValueError: Cannot assign "1": "SiteDailyMetrics.site" must be a
+            "Site" instance
+
+        Since we do want to constrain uniqueness per site+day, we'll need to fix
+        this
+        """
+        ordering = ['-date_for', 'site']
 
     def __str__(self):
-        return "{} {}".format(self.id, self.date_for)
+        return "id:{}, date_for:{}, site:{}".format(
+            self.id, self.date_for, self.site.domain)
 
 
 class LearnerCourseGradeMetricsManager(models.Manager):
@@ -102,9 +117,10 @@ class LearnerCourseGradeMetrics(TimeStampedModel):
 
     """
     date_for = models.DateField()
+    # TODO: We should require the user
     user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
     course_id = models.CharField(max_length=255, blank=True)
-
+    site = models.ForeignKey(Site, default=settings.SITE_ID)
     points_possible = models.FloatField()
     points_earned = models.FloatField()
     sections_worked = models.IntegerField()
@@ -170,6 +186,7 @@ class PipelineError(TimeStampedModel):
     # Attributes for convenient querying
     course_id = models.CharField(max_length=255, blank=True)
     user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
+    site = models.ForeignKey(Site, blank=True, null=True)
 
     class Meta:
         ordering = ['-created']
--- a/figures/models.py
+++ b/figures/models.py
@@ -21,6 +21,7 @@ class CourseDailyMetrics(TimeStampedModel):
     adding a SiteDailyMetrics foreign key. This is subject to change as the code
     evolves.
     """
+    site = models.ForeignKey(Site)
     date_for = models.DateField()
 
     # Leaving as a simple string for initial development
@@ -28,7 +29,6 @@ class CourseDailyMetrics(TimeStampedModel):
     # the CourseOverview model or have the course_id be a
     # CourseKeyField
     course_id = models.CharField(max_length=255)
-    site = models.ForeignKey(Site, default=settings.SITE_ID)
     enrollment_count = models.IntegerField()
     active_learners_today = models.IntegerField()
     # Do we want cumulative average progress for the month?
@@ -54,9 +54,9 @@ class SiteDailyMetrics(TimeStampedModel):
     Stores metrics for a given site and day
     """
 
+    site = models.ForeignKey(Site)
     # Date for which this record's data are collected
     date_for = models.DateField()
-    site = models.ForeignKey(Site, default=settings.SITE_ID)
     cumulative_active_user_count = models.IntegerField(blank=True, null=True)
     todays_active_user_count = models.IntegerField(blank=True, null=True)
     total_user_count = models.IntegerField()
@@ -116,11 +116,11 @@ class LearnerCourseGradeMetrics(TimeStampedModel):
     actually needed and edx-platform uses FloatField in its grades models
 
     """
+    site = models.ForeignKey(Site)
     date_for = models.DateField()
     # TODO: We should require the user
     user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
     course_id = models.CharField(max_length=255, blank=True)
-    site = models.ForeignKey(Site, default=settings.SITE_ID)
     points_possible = models.FloatField()
     points_earned = models.FloatField()
     sections_worked = models.IntegerField()
--- a/figures/permissions.py
+++ b/figures/permissions.py
@@ -10,13 +10,46 @@ import organizations
 import figures.settings
 
 
-def is_site_admin_user(request):
+class MultipleOrgsPerUserNotSupported(Exception):
+    pass
 
-    if not figures.settings.is_multisite():
-        has_permission = IsStaffUser().has_permission(request, None)
-    else:
+
+def is_site_admin_user(request):
+    """
+    Determines if the requesting user has access to site admin data
+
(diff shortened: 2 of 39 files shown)


johnbaldwin:
[(2) Add multi-site settings to Figures, tests](https://trello.com/c/ZBmrWx4y/681-2-add-multi-site-settings-to-figures-tests)

johnbaldwin:
[(3) Add Site to metrics models, tests](https://trello.com/c/pk75I1Bk/684-3-add-site-to-metrics-models-tests)

tests/test_settings.py
```diff
@@ -13,6 +13,25 @@
...
 
+@pytest.mark.parametrize('env_tokens, expected ', [
+        ({'IS_MULTISITE': True}, True),
+        ({'IS_MULTISITE': False}, False),
+    ])
+def test_is_multisite(env_tokens, expected):
+    """
+    Test env_tokens work properly for Figures multisite settings
+    """
+    with mock.patch('figures.settings.env_tokens', env_tokens):
```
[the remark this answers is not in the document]
johnbaldwin: I'm hesitant to use `override_settings` because it looks like `override_settings` does not support nested keys and at this time: https://github.com/django/django/blob/1.8.18/django/test/utils.py#L221-L223

I'd rather just touch the Figures nested dict within `ENV_TOKENS` than have to make sure that each test that overrides settings has properly crafted `ENV_TOKENS`. Even though "right now" `ENV_TOKENS` in the Figures `devsite/test_settings.py just defines `FIGURES` as its only key, there is no guarantee this will stay, in which case `override_settings` becomes a problem.

At this time, I also want to avoid spending the time to do yet another rewrite of how Figures settings are retrieved from edx-platform. I'm happy to revisit when we work on the Hawthorn upgrade. I think it would really help if I do a sequence diagram that illustrates the Figures app instantiating. 

After digging through the `override_settings` code, it appears there may be more robust avenue by using signals. But top priority is getting Figures working in a multisite environment

figures/settings.py
```diff
@@ -15,11 +15,20 @@
...
 DEFAULT_LOG_PIPELINE_ERRORS_TO_DB = True
 
 env_tokens = {}
 
 
+def is_multisite():
+    """
+    Override by setting ``IS_MULTISITE`` to true in the Figures env settings
+    """
+    return bool(env_tokens.get('IS_MULTISITE',
```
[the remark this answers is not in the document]
johnbaldwin: Since this method is not updating `figures.settings.env_tokens`, the update you suggest will work. I'll do this

tests/test_settings.py
```diff
@@ -13,6 +13,25 @@
 from figures import settings as figures_settings
 
 
+@pytest.mark.parametrize('env_tokens, expected ', [
+        ({'IS_MULTISITE': True}, True),
+        ({'IS_MULTISITE': False}, False),
```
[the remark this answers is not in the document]
johnbaldwin: I think so. Right now I don't see a reason why this won't work. Thanks, I'll try that!

figures/settings.py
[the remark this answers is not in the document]
johnbaldwin: @OmarIthawi Are you suggesting doing a lazy import within the `is_multisite` function?

tests/test_settings.py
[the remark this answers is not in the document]
OmarIthawi: yeah, can't help with the future

figures/settings.py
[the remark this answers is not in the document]
johnbaldwin: @OmarIthawi Bump on this question, please.

johnbaldwin:
I added tox support. It's a bit of a hack and could use cleaning up, but works well enough right now that I don't want to put time into it now. So I created an issue to track tox tech debt: https://github.com/appsembler/figures/issues/69

figures/models.py
```diff
@@ -23,7 +28,7 @@ class CourseDailyMetrics(TimeStampedModel):
     # the CourseOverview model or have the course_id be a
     # CourseKeyField
     course_id = models.CharField(max_length=255)
-
+    site = models.ForeignKey(Site, default=settings.SITE_ID)
```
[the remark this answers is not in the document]
OmarIthawi: That would be better. Thank @johnbaldwin!

figures/settings.py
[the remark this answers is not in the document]
OmarIthawi: @johnbaldwin I think the idea of a plugin patching `settings` is very buggy and involves guessing on which module was loaded first. Instead, whenever possible, `settings` should be loaded as late as reasonably possible, and only read and not written.

What I'm suggesting is not just lazy loading. It is lazy loading to avoid Django getting confused with partially initialized settings, since the `settings` also imports a function from Figures.

 - https://github.com/appsembler/edx-platform/blob/bd259531edaec0e8824349ec1a7d20e75317e3e2/lms/envs/aws_appsembler.py#L155-L161

I'm also suggesting avoiding writing on `settings` and `ENV_TOKENS` except when absolutely needed.

There's a notable exception for the above which is the Hawthorn plugin system, but that would be out of the scope of this PR.

Please let me know if you have any questions.

johnbaldwin:
[(5) Add multi-site permissions to Figures permissions module, tests](https://trello.com/c/Li3zhp5i/682-5-add-multi-site-permissions-to-figures-permissions-module-tests)

figures/permissions.py
```diff
@@ -3,11 +3,91 @@
...
 
+import django.contrib.sites.shortcuts
 
-class IsStaffUser(BasePermission):
+import organizations
+
+import figures.settings
+
+
+class MultipleOrgsPerUserNotSupported(Exception):
```
OmarIthawi: Someone will get this error, sooner or later. It helps to add some docs here to specify why this is not supported, and if there's support.

OmarIthawi: nvm, found the why :point_down:

figures/serializers.py
```diff
@@ -335,6 +349,7 @@ def get_learners_enrolled(self, course_overview):
         linked
         """
         return get_course_history_metric(
+            site=self.site,
```
OmarIthawi: Why the hack above for `to_representation`, why not do the below instead:

```suggestion
            site=figures.sites.get_site_for_course(course_overview),
```

figures/sites.py
```diff
@@ -0,0 +1,126 @@
...
+
+def get_course_enrollments_for_site(site):
+    course_keys = get_course_keys_for_site(site)
+    return CourseEnrollment.objects.filter(course_id__in=course_keys)
+
+
+# MultipleObjectsReturned: get() returned more than one UserOrganizationMapping -- it returned 5!
+
+#  File "/edx/src/figures/figures/orgs.py", line 40, in get_users_for_site
+#    organization__in=orgs)
```
OmarIthawi: Please delete the comments above.

figures/tasks.py
```diff
@@ -91,12 +93,19 @@ def populate_daily_metrics(date_for=None, force_update=False):
     logger.info('Starting task "figures.populate_daily_metrics" for date "{}"'.format(
         date_for))
 
-    for course in CourseOverview.objects.all():
-        populate_single_cdm(
-            course_id=course.id,
+    # print('testing...')
+    # import pdb; pdb.set_trace()
```
OmarIthawi: remove plz

figures/views.py
```diff
@@ -166,28 +178,29 @@ def get_queryset(self):
 class CourseDailyMetricsViewSet(CommonAuthMixin, viewsets.ModelViewSet):
 
     model = CourseDailyMetrics
-    queryset = CourseDailyMetrics.objects.all()
+    # queryset = CourseDailyMetrics.objects.all()
```
OmarIthawi: please delete.

melvinsoft:
@johnbaldwin Thanks! Reviewing this PR today.

figures/tasks.py
```diff
@@ -17,7 +18,7 @@
 from figures.helpers import as_course_key, as_date
 from figures.pipeline.course_daily_metrics import CourseDailyMetricsLoader
 from figures.pipeline.site_daily_metrics import SiteDailyMetricsLoader
-
+import figures.sites
```
melvinsoft: @johnbaldwin Why do you prefer to import the whole `figures.sites` helper rather than importing just the function that is gonna be used here?

tests/factories.py
```diff
@@ -11,27 +11,55 @@
...
+
 import factory
 from factory import fuzzy
 from factory.django import DjangoModelFactory
 
 from openedx.core.djangoapps.content.course_overviews.models import (
     CourseOverview,
 )
-
+# from openedx.core.djangoapps.site_configuration.models import SiteConfiguration
```
melvinsoft: @johnbaldwin Remove it please.

tests/factories.py
```diff
@@ -11,27 +11,55 @@
...
 from openedx.core.djangoapps.xmodule_django.models import CourseKeyField
 from certificates.models import GeneratedCertificate
 from courseware.models import StudentModule
 from student.models import CourseAccessRole, CourseEnrollment, UserProfile
 from lms.djangoapps.teams.models import CourseTeam, CourseTeamMembership
 
+import organizations
+# from organizations import tests as org_tests
```
melvinsoft: @johnbaldwin Remove it please.

tests/factories.py
```diff
@@ -11,27 +11,55 @@
...
 from figures.models import CourseDailyMetrics, SiteDailyMetrics
 
+from tests.helpers import organizations_support_sites
+
+
 COURSE_ID_STR_TEMPLATE = 'course-v1:StarFleetAcademy+SFA{}+2161'
 
 
+
+# class SiteConfigurationFactory(factory.DjangoModelFactory):
```
melvinsoft: @johnbaldwin Please remove if we don't need it.

johnbaldwin:
[(8) Add multi-site mode to Figures views, tests](https://trello.com/c/RlnDKqar/683-8-add-multi-site-mode-to-figures-views-tests)

figures/sites.py
johnbaldwin: ok

figures/views.py
johnbaldwin: Yep

figures/tasks.py
johnbaldwin: @melvinsoft I learned it is easier to use `mock` when the test class and the production code use the same module name so that when the production code is called from test that uses `figures.sites` (or any full namespace for that matter), then you can use `mock.patch`

Example:
```
with mock.patch('figures.settings.env_tokens', self.env_tokens):
    do_something(x)
```

When I do `from x import y` the production code doesn't seem consistent in using the mocked value. There are a number of examples on the web mentioning this. Here's one SO: https://stackoverflow.com/questions/11351382/mock-patching-from-import-statement-in-python

Let me know if I need to explain this more

tests/factories.py
johnbaldwin: yep, will do

tests/factories.py
johnbaldwin: will do

tests/factories.py
johnbaldwin: will do

johnbaldwin:
@OmarIthawi @melvinsoft Thank you very much for the review comments!

figures/permissions.py
```diff
@@ -3,11 +3,91 @@
...
+                'Only one org per user per site is allowed. found {}'.format(
+                    uom_qs.count())
+                )
+        # Since Tahoe does just one org, we're going to cheat and just look
+        # for the first element
+        if uom_qs and request.user.is_active:
+            has_permission = uom_qs[0].is_amc_admin and uom_qs[0].is_active
+        else:
+            has_permission = False
+    else:
```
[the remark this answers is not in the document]
johnbaldwin: @OmarIthawi Are you running into an issue with permissions as is? It should pass the tests/test_permissions.py tests. I'd prefer not to have multiple exit points in functions unless there's a compelling reason to do so. For reference: https://docs.python-guide.org/writing/style/#returning-values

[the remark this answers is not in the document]
OmarIthawi: Never heard about it before. Thanks @johnbaldwin!

figures/tasks.py
melvinsoft: @johnbaldwin Thanks for the explanation! Makes sense.

OmarIthawi: @johnbaldwin for your reference, it you'd like to do an `from import figures.sites import funky_function` in `figures/tasks.py`.

You should patch it like below:

```
with mock.patch('figures.tasks.funky_function`, self.funky_function):
  do_something(x)
```

It's confusing, but this is how it works.

PR #72, opened 2019-02-05
fix learner counts

johnbaldwin:
This PR replaces direct calls to `CourseEnrollment.objects.filter` and replaces with calls to `get_enrolled_in_exclude_admins` in figures.pipeline.course_daily_metrics.

Made the `date_for` parameter optional in `get_enrolled_in_exclude_admins`  to make the code cleaner. 

I pushed to staging and manually ran the pipeline succesfully

### DIFF
--- a/figures/filters.py
+++ b/figures/filters.py
@@ -12,6 +12,7 @@ from openedx.core.djangoapps.content.course_overviews.models import (
 )
 from student.models import CourseEnrollment
 
+from figures.pipeline.course_daily_metrics import get_enrolled_in_exclude_admins
 from figures.models import CourseDailyMetrics, SiteDailyMetrics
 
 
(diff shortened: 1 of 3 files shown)


figures/pipeline/course_daily_metrics.py
```diff
@@ -56,7 +56,7 @@ def get_num_enrolled_in_exclude_admins(course_id, date_for):
     and modified to filter on date LT
 
     """
-    course_locator = course_id
+    course_locator = as_course_key(course_id)
```
johnbaldwin: This wasn't causing a failure, but is a latent bug, so fixed it

johnbaldwin:
[Fix metrics to exclude course admins/staff from stats](https://trello.com/c/aNvgdSrH/724-fix-metrics-to-exclude-course-admins-staff-from-stats)

figures/filters.py
```diff
@@ -12,6 +12,7 @@
 )
 from student.models import CourseEnrollment
 
+from figures.pipeline.course_daily_metrics import get_enrolled_in_exclude_admins
```
johnbaldwin: Ideally we want a different home module for this function. I'd put it in sites, but the course id is what drives the filtering. I'll likely revisit this when I start looking into object level permissions

johnbaldwin:
@OmarIthawi @melvinsoft Awesome, thanks guys! Merged and I'll cut a tag ASAP today

PR #73, opened 2019-02-13
Allow staff and superusers to access figures for any site

johnbaldwin:
This PR is to allow staff and superusers to access figures for any site in a multisite environment. 

The `is_site_admin_user` method has been updated to first check if the user has staff or superuser permissions. If not, then execute the multisite check (which should be unchanged, just indented).

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -76,10 +76,15 @@ class TestSiteAdminPermissionsForMultisiteMode(object):
                 organization=self.organization,
                 is_amc_admin=True)
         ]
+        self.callers += create_test_users()
         self.env_tokens = {'IS_FIGURES_MULTISITE': True}
 
     # @mock.patch('figures.settings')
     @pytest.mark.parametrize('username, allow', [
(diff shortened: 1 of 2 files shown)


johnbaldwin:
It seems that the coverage is failing because codecov is not running the multisite configuration. `figures.permissions` has 100% coverage when run using the multisite environment (https://github.com/appsembler/figures/blob/develop/tox.ini#L9)

tests/test_permissions.py
```diff
@@ -97,7 +102,7 @@ def test_site(request):
 
             permission = figures.permissions.IsSiteAdminUser().has_permission(
                 request, None)
-            assert permission == allow
+            assert permission is allow, 'user {}'.format(username)
```
OmarIthawi: Nitpick

```suggestion
            assert permission is allow, 'User "{username}" should have access'.format(username=username)
```

johnbaldwin: Sure, I'll indulge in your nitpick

PR #74, opened 2019-02-19
Feature/read data values with fallback

grozdanowski:
These changes fix the issue that happens when the data that the API sends is not exactly as the app expects. We utilise Immutable and its .getIn() method which doesn't break if some nested values don't exist.

Note: this PR doesn't address error reporting/handling, it just makes the app not break when data structure is not right. Proper error reporting/handling will be part of another PR.


frontend/src/views/DashboardContent.js
```diff
@@ -43,24 +44,24 @@ class DashboardContent extends Component {
         </HeaderAreaLayout>
         <div className={cx({ 'container': true, 'base-grid-layout': true, 'dashboard-content': true})}>
           <BaseStatCard
-            cardTitle='Number of registered learners'
-            mainValue={this.props.generalData['total_site_users']['current_month']}
-            valueHistory={this.props.generalData['total_site_users']['history']}
+            cardTitle='Number of registered learners asadadad'
```
johnbaldwin: @grozdanowski What's 'asadadad'?

PR #78, opened 2019-03-01
Fix staff name display in Dashboard

grozdanowski:
Staff names weren't being pulled in correctly (wasn't using Immutable .get function).


johnbaldwin:
LGTM!

PR #80, opened 2019-03-16
Bug fixes after feedback from production push

grozdanowski:
These commits take care of the following bugs:

- lots of decimals in course progress display
- dates not formatted right when displayed or not displaying at all
- add +/- sign in front of delta of MAU's in MAU's by month table
- nomenclature change - "Dashboard" -> "Overview"
- average course progress displayed as a 0.x value, when it should display a percentage
- users distribution across course section displaying wrong data - had to simply remove this graph. Turns out the backend still doesn't have an API endpoint that exposes this data.
- weird color of user names in users list in single course page
- UX issue - make the entire course card in course list clickable
- add tooltips that display actual data to the graph that displays MAU's progress (header area)
- fix an issue where "See history" displays a white graph that when hovered breaks the app (data type issue)
- learner statistics graph not working - now it works, and loads straight away.
- small accessibility issue - "Jump to content" accessibility link didn't work. Now it does.
- nomenclature of card titles in Dashboard/Overview
- **set default SASS variables to Appsembler-preferred** - we're doing this so the app plays great with Appsembler's theme codebase out of the box. In order for the app to display correctly parameters need to be passed when rebuilding code (to have correct font sizes: `REM_BASE=16 yarn build`)


johnbaldwin:
@grozdanowski I just skimmed over. Will look over in more depth tomorrow. One thing we want to make sure is clear is when and how the custom yarn build needs to be done. 

What command do you run to build the assets we will commit to a release? I want to add that to the makefile.

frontend/src/views/MauDetailsContent.js
```diff
@@ -19,7 +19,7 @@ class MauDetailsContent extends Component {
         <li key={index} className={styles['content-row']}>
           <span className={styles['period']}>{period.period}</span>
           <span className={styles['mau-count']}>{period.value}</span>
-          <span className={cx({ 'difference': true, 'positive': (difference > 0), 'negative': (difference < 0)})}>{difference}</span>
+          <span className={cx({ 'difference': true, 'positive': (difference > 0), 'negative': (difference < 0)})}>{(difference > 0) ? "+" : ""}{difference}</span>
```
OmarIthawi: Can we make this sign an emoji like :arrow_up_small: and :arrow_down_small:

frontend/src/components/header-views/header-content-course/HeaderContentCourse.js
```diff
@@ -39,6 +39,8 @@ class HeaderContentCourse extends Component {
 
   render() {
 
+    const displayCourseHeaderGraph = false;
```
OmarIthawi: Yes please! The graph is broken in my opinion, thanks for removing it!

I should've reported the bug earlier, but for whatever reason I didn't.

frontend/src/components/courses-list/CoursesListItem.js
```diff
@@ -24,7 +24,7 @@ class CoursesListItem extends Component {
     });
 
     return (
-      <div className={styles['course-list-item']} key={this.props.courseId}>
+      <Link to={'/figures/course/' + this.props.courseId} className={styles['course-list-item']} key={this.props.courseId}>
```
OmarIthawi: :+1:

frontend/src/App.js
```diff
@@ -26,7 +26,7 @@ class App extends Component {
       <LoadingSpinner
         displaySpinner = {!(this.props.activeApiFetches === 0)}
       >
-        <main className={styles['layout-root']}>
+        <main id="main" className={styles['layout-root']}>
```
OmarIthawi: I'm sure @mhaton is happier now :smiley:

frontend/src/sass/base/_variables.scss
```diff
@@ -1,14 +1,14 @@
 @import '~base/sass/base/functions';
 
-$rem-base-size: 16 !default;
+$rem-base-size: 10 !default;
```
OmarIthawi: Will this break when the community uses them on vanilla Open edX?

johnbaldwin: @OmarIthawi If this does break for vanilla, I'll work out an interim solution. Not for now, but for later, is there a reason why an external css that defines the `rem-base-size`? from which Figures uses. I mean, this is part of the cascade in CSS, right? That and we really want to move away from building the assets and including them in the repo, instead building the assets into the deployable python bundle we push to PyPI

until we automate our build/deploy to PyPI and th

johnbaldwin:
@grozdanowski Thanks for jamming this out this weekend!

frontend/src/sass/base/_variables.scss
OmarIthawi: Thanks @johnbaldwin! I did use the wrong word here. I don't think it will "break break", and [Matej explained that this is a temporary solution](https://appsembler.slack.com/archives/C4FHDM4U9/p1552848244002800?thread_ts=1552757366.001000&cid=C4FHDM4U9), which I think is OK.

I simply worried that the font will look very small on Open edX deployments to the point that forces community members to fork and recompile, which is anti-opensource.

> we really want to move away from building the assets and including them in the repo, instead building the assets into the deployable python bundle we push to PyPI

Yes please! I really like that you're working on keeping PyPI as the no. 1 installation option.

> If this does break for vanilla, I'll work out an interim solution. Not for now, but for later, is there a reason why an external css that defines the rem-base-size?

Yes, `rem` is all about cascade/inheritance as far as I know. @grozdanowski knows better than I do, but I think it's more like saying `font-size: 100%`. It's not just cascaded, it is also depends on what the parent `font-size` value is.

OmarIthawi:
@grozdanowski I took the liberty to reword the title of this pull request :see_no_evil:

frontend/src/views/MauDetailsContent.js
grozdanowski: @OmarIthawi we can, but I'd rather do it in FontAwesome than (ugly) emoji :) Let's keep this off for now - I don't think it's a priority.

frontend/src/sass/base/_variables.scss
grozdanowski: @johnbaldwin @OmarIthawi To clarify - there's no need to fork. One can pass on env variables and rebuild on the fly. This is also a convenient tool for other purposes (if someone wants to brand their Figures intance for themselves using font sizes or colours).
And yes - it is an issue in regards to cascading nature of CSS - `rem` is defined by the first ever definition of font size in the rendered document.

grozdanowski:
Thanks for reviewing this - merging.

PR #81, opened 2019-03-21
Backend fixes from BugFest

johnbaldwin:
## Cross-site courses

Now only courses for the current site should show in the course list on the user page.

This one is a bit interesting. The learner details serializer is required to provide a list of the courses for which the learner is enrolled. Previously it was not filtering courses on the current site. To make the learner details serializer course list filter on site, the site object or a site unique identifier needs to be available to the serializer's `get_courses` method. One option I looked into was adding `site` to the serializer constructors, but it felt like a hack, so I investigated and learned about the `context` 
argument: https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context.

`LearnerDetailsViewSet` now has a `retrieve` method that passes the current site as the context to the serializer.

## Course start dates

Should no longer start at "Thu, 01 Jan 1970 00:00:00 GMT"

Figures was using `CourseOverview` `enrollment_start` and `enrollment_end` for the course start and end dates. Now `CourseOvervew` `start` and `end` fields are used.

## Learner List

Now learners list should not be empty if the course has enrolled learners.

The problem was that the `action` parameter in the UserFilterSet method filter field `enrolled_in_course_id`. It had the `filter_` prefix.

The front end calls `/figures/api/learners/details/?enrolled_in_course_id` but the filter was expecting `/figures/api/learners/details/?filter_enrolled_in_course_id`.

Also fixed are the other filters that used the `filter_` prefix in the action parameter.

## Testing

Updated tests for the code changes and added a test stub for excluding other sites. I created a card to follow up with this

### DIFF
--- a/figures/filters.py
+++ b/figures/filters.py
@@ -53,7 +53,7 @@ class CourseEnrollmentFilter(django_filters.FilterSet):
 
     '''
 
-    course_id = django_filters.MethodFilter(action='filter_course_id')
+    course_id = django_filters.MethodFilter(action='course_id')
     is_active = django_filters.BooleanFilter(name='is_active',)
 
     def filter_course_id(self, queryset, course_id_str):
@@ -84,15 +84,17 @@ class UserFilterSet(django_filters.FilterSet):
     We're starting with a few fields and will add as we find we want/need them
 
     '''
-    is_active = django_filters.BooleanFilter(name='is_active',)
+    is_active = django_filters.BooleanFilter(name='is_active')
+    is_staff = django_filters.BooleanFilter(name='is_staff')
+    is_superuser = django_filters.BooleanFilter(name='is_superuser')
     username = django_filters.CharFilter(lookup_type='icontains')
     email = django_filters.CharFilter(lookup_type='icontains')
     country = django_filters.CharFilter(
         name='profile__country', lookup_type='iexact')
 
-    user_ids = django_filters.MethodFilter(action='filter_user_ids')
+    user_ids = django_filters.MethodFilter(action='user_ids')
     enrolled_in_course_id = django_filters.MethodFilter(
-        action='filter_enrolled_in_course_id')
+        action='enrolled_in_course_id')
 
     class Meta:
         model = get_user_model()
--- a/figures/pipeline/course_daily_metrics.py
+++ b/figures/pipeline/course_daily_metrics.py
@@ -55,6 +55,8 @@ def get_enrolled_in_exclude_admins(course_id, date_for=None):
     Copied over from CourseEnrollmentManager.num_enrolled_in_exclude_admins method
     and modified to filter on date LT
 
+    If no date is provided then the date is not used as a filter
+
     """
     course_locator = as_course_key(course_id)
 
@@ -243,7 +245,7 @@ class CourseDailyMetricsExtractor(object):
         # set of calls defined in a ruleset instead of hardcoded here after
         # retrieving the core quersets
 
-        course_enrollments = get_course_enrollments(
+        course_enrollments = get_enrolled_in_exclude_admins(
             course_id, date_for,)
 
         data = dict(date_for=date_for, course_id=course_id)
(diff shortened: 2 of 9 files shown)


figures/filters.py
```diff
@@ -84,15 +84,17 @@ class UserFilterSet(django_filters.FilterSet):
     We're starting with a few fields and will add as we find we want/need them
 
     '''
-    is_active = django_filters.BooleanFilter(name='is_active',)
+    is_active = django_filters.BooleanFilter(name='is_active')
+    is_staff = django_filters.BooleanFilter(name='is_staff')
+    is_superuser = django_filters.BooleanFilter(name='is_superuser')
```
johnbaldwin: I took the opportunity to add `is_staff` and `is_superuser` filters

figures/pipeline/course_daily_metrics.py
```diff
@@ -35,26 +35,16 @@
 logger = logging.getLogger(__name__)
 
 
-#
 # Extraction helper methods
-#
-
-
-def get_course_enrollments(course_id, date_for):
```
johnbaldwin: Method no longer needed

figures/pipeline/course_daily_metrics.py
```diff
@@ -243,7 +233,7 @@ def extract(self, course_id, date_for=None, **kwargs):
         # set of calls defined in a ruleset instead of hardcoded here after
         # retrieving the core quersets
 
-        course_enrollments = get_course_enrollments(
+        course_enrollments = get_enrolled_in_exclude_admins(
```
johnbaldwin: small optimization to remove an unnecessary query since our context is that only learners should be retrieved from `CourseEnrollment`

figures/serializers.py
```diff
@@ -715,9 +715,12 @@ def get_courses(self, user):
         """
         This method is a hack until I figure out customizing DRF fields and/or
         related serializers to explicitly link models not linked via FK
+
         """
-        return LearnerCourseDetailsSerializer(
-            CourseEnrollment.objects.filter(user=user), many=True).data
+
+        course_enrollments = figures.sites.get_course_enrollments_for_site(
+            self.context.get('site')).filter(user=user)
```
johnbaldwin: We now use DRF serializers `context` construct to provide the `site` model to the sites method

figures/views.py
```diff
@@ -272,21 +272,19 @@ class CourseDetailsViewSet(CommonAuthMixin, viewsets.ReadOnlyModelViewSet):
...
     def get_queryset(self):
-        print('CourseDetailsViewSet.get_queryset called')
         site = django.contrib.sites.shortcuts.get_current_site(self.request)
         queryset = figures.sites.get_courses_for_site(site)
         return queryset
 
     def retrieve(self, request, *args, **kwargs):
-        print('CourseDetailsSerializer.retrieve called')
         # NOTE: Duplicating code in GeneralCourseDataViewSet. Candidate to dry up
+        # Make it a decorator
```
johnbaldwin: Just a note here since I'm in views a lot. retrieving the site from the course id string is a perfect candidate for a decorator

tests/test_serializers.py
```diff
@@ -455,6 +458,13 @@ def test_has_fields(self):
         # model (UserProfile) data
         assert data['name'] == 'Alpha One'
 
+    @pytest.mark.skip(reason='Not implemented')
```
johnbaldwin: I created a card to implement this. This one is a victim of schedule

figures/serializers.py
abeals: what happens here if `site` isn't in the `context` dict?

johnbaldwin: Good catch! This is a latent bug. The LearnerDetailsSerializer is only used in LearnerDetailsViewSet, so for the current state of the code, context always has a `site` key. I'll create a ticket to track and fix this.

johnbaldwin: Good catch! This is a latent bug. Currently, LearnerDetailsSerializer is only called by LearnerDetailsViewSet. Created a ticket to track and address this: https://github.com/appsembler/figures/issues/87

PR #105, opened 2019-07-27
Updated Figures Makefile to simplify updating mock models migrations

johnbaldwin:
Background: Django 1.9 removed supported for syncing apps without migrations per deprecation timeline #3791

See here: https://github.com/django/django/pull/3791

This PR actually fixes a nonworking target I had to update the mocks' migrations

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 # Work in progress makefile to simplify Figures development
 #
 
-.PHONY: help clean_tests clean_webpack_buld clean_python_build build_python lint pip_install test twine_check twine_push_test
+.PHONY: help clean_tests clean_webpack_buld clean_python_build build_python lint pip_install test twine_check twine_push_test delete_mock_migrations reset_mock_migrations
 
 help:
 	@echo "Targets:"
@@ -14,9 +14,11 @@ help:
 	@echo " - lint"
 	@echo " - pip_install"
 	@echo " - test"
-	@echo " -twine_check"
-	@echo " -twine_push_test"
-	@echo " -twine_push_prod"
+	@echo " - twine_check"
+	@echo " - twine_push_test"
+	@echo " - twine_push_prod"
+	@echo " - delete_mock_migrations"
+	@echo " - update_mock_migrations"
 
 clean_python:
 	find tests -type f -name "*.pyc" -exec rm -f {} \;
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,9 +14,11 @@ help:
 	@echo " - lint"
 	@echo " - pip_install"
 	@echo " - test"
-	@echo " -twine_check"
-	@echo " -twine_push_test"
-	@echo " -twine_push_prod"
```
johnbaldwin: Also a little cleanup

OmarIthawi:
@johnbaldwin I checked the Django pull request, but couldn't really understand what does it have to do with this PR.

I don't see you're using the old `manage.py sync` command.

johnbaldwin:
@OmarIthawi I don't need a more detailed review, but thanks for asking.  I'll explain a bit more:

Back in Django 1.8 (Open edX Ginkgo), you could define a model without requiring a migration for the model. Obviously, you don't want to to that in production, but for tests where the database is created and then discarded after the test, this worked great. However once we upgraded to Hawthorn, we use Django 1.11 and this feature of "migration-less models" is no longer possible because it was removed in Django 1.9 (the Django PR link explains this). So now we need migrations for the mock models. We still just create, use, and discard the models, so there's no real persistence. Therefore, we don't need to incrementally update models for changes. We just need to define the model in the migration for the model "right now". So the makefile targets I updated here help automate the work of deleting the current mock migrations and recreate them.

Yeah, it's a hack, but I think it an acceptable hack because otherwise, we'd have to choose between

A) Mock the edx-platform models with mocks/monkeypatches, which can get really ugly fast unless we're really careful about it. Time I don't think we have the luxury to spend.

B) Inherit and modernize a "Fake Model" library like this one: https://github.com/erm0l0v/django-fake-model. Which again gets into spending time we don't have the luxury to spend.

Note, this is NOT the same as FactoryBoy, which requires something that acts as a real model that can persist to the test database

C) Do away with mocking edx-platform in Figures, which means we need to test Figures only in an edx-platofrm environment. This is unacceptable as it would turn unit tests that now run in seconds to minutes or worse.

Now for a long term solution: Abstract Figures so that it becomes a more generic reporting/analytics platform. We define an interface to define data sets, then map those definitions to implementation specific data models. Then we can have an `edx-platform` pipeline backend plugin that IS only tested in edx-platform, doing away with `edx-platform` mocks in Figures. I'll start working on a TDP about this as I've got some ideas for Python metaprogramming that can make it happen. The big unknown to me right now is getting the front end abstracted. I "just" have to play with the front end a bit in like 2-3 spikes to build a front end reporting abstraction layer.

OmarEdraak:
Makes a lot of sense now @johnbaldwin! Thanks for explaining that!

OmarIthawi:
Makes sense now! Thanks @johnbaldwin for the detailed explanation.

PR #113, opened 2019-10-09
Fix dev styles, improve responsiveness

grozdanowski:
In dev mode (standalone Django dev app) everything was oversized and paddings were non-existing. That's fixed with this PR.
Also I added some fixes in responsiveness of the Dashboard view.


johnbaldwin:
@grozdanowski I'm merging this in

PR #114, opened 2019-10-12
Test fixes - removed local hardcoded date string from view tests

johnbaldwin:
Removed hard-coded local scope variables because they overwrote the parametrize declared test variables


johnbaldwin:
Thanks @OmarIthawi !

PR #115, opened 2019-10-23
Incluide CELERY_IMPORTS in settings

melvinsoft:
remove changes in tasks


OmarIthawi:
So that's it!! That would be awesome!

PR #120, opened 2019-10-23
Devsite - fixed failing 'seed_data' command

johnbaldwin:


johnbaldwin:
Thanks @melvinsoft 

The error is that the `settings.FEATURES` is missing from the devsite settings. `FEATURES` is present in `test_settings.py` so tests work, but devsite breaks because `FEATURES` is missing. Is this enough context?

As for the failing test, this is test coverage, not actual tests. Test coverage is being reported as reduced, but I'm not sure why a one line change in the settings would reduce test coverage. Since this is for devsite/demo mode, I'm not concerned about it. It has no impact in production. Does that help explain this?

melvinsoft:
John! Awesome! Thanks for the clear explanation here!

PR #122, opened 2019-10-29
API Hack: allow up to 1000 general course data records to be returned

johnbaldwin:
We are doing this so that we don't have to implement pagination for the
front end site page. This is a tempoarary hack that we need to fix as we
revisit the API

@OmarIthawi This was an agreed upon hack between Matej and John. We know we need to revisit the API design

### DIFF
--- a/figures/views.py
+++ b/figures/views.py
@@ -52,7 +52,10 @@ from figures.serializers import (
     GeneralUserDataSerializer
 )
 from figures import metrics
-from figures.pagination import FiguresLimitOffsetPagination
+from figures.pagination import (
+    FiguresLimitOffsetPagination,
+    FiguresKiloPagination,
+)
 import figures.permissions
 import figures.helpers
 import figures.sites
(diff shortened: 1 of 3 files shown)


figures/views.py
```diff
@@ -263,7 +263,10 @@ class GeneralCourseDataViewSet(CommonAuthMixin, viewsets.ReadOnlyModelViewSet):
 
     '''
     model = CourseOverview
-    pagination_class = FiguresLimitOffsetPagination
+
+    # The "kilo paginator"  is a tempoarary hack to return all course to not
+    # have to change the front end until Figures "Level 2"
+    pagination_class = FiguresKiloPagination
```
thraxil: FYI, `FiguresKiloPagination` is not imported in `views.py`, so this should be a runtime error. Flake8 correctly points it out, which is why the TravisCI tests are failing.

johnbaldwin: @thraxil Thank you very much! Fixed an added test

johnbaldwin:
Add pagination with 1000 records for the general course data viewset as a temporary fix for the course pagination issue

PR #136, opened 2019-11-22
WIP baseline test and mock changes to support Ginkgo and Hawthorn

johnbaldwin:
This is the initial PR to provide Ginkgo backport support. 

The purpose is to lay out multi-version Open edX support into common code before adding Ginkgo specific changes to the production code and tests

We're changing the mocks layout by removing `mocks` from the `tests` directory, promoting to the project top level, `<figures project root>/mocks`. Within `mocks` will be the mock edx-platform code for each named Open edX release.

Then we make Pytest and Tox use the Hawthorn mocks. This allows us to follow on with Ginkgo specifics in the production code and tests. We may still need to have a Ginkgo specific branch, but at least we will have less difference in the branches


johnbaldwin:
Closing PR and deleting branch. Will use a different approach for the series of PRs

PR #137, opened 2019-11-22
Restructured mocks to support multiple edx-platform mocking

johnbaldwin:
What:

- Moved mocks from `./tests/mocks` to `./mocks/hawthorn`
- Moved mocks sys.path inclusion from manage.py to devsite/test_settings

Why:

This is to support mocking multiple Open edX versions testing with the
same codebase. Ideally we want to be able to support multiple platforms
with the same branch. If we cannot achieve that goal, we can at least
minimze the differences between the branches making maintenance less
costly


johnbaldwin:
@OmarIthawi Thanks for the approval!

PR #138, opened 2019-11-23
Fixes getting previously collected days data when 'yesterday' was not collected

johnbaldwin:
Please see the git log for each commit. The first commit is important for finding the last record saved even if not yesterday. The second commit uses this new functionality to retrieve previous data for the site daily metrics.

As commented in the second commit, It does does NOT fix the issue that the SiteDailyMetrics active user count has flawed data. The flaw is that it collects the totals from the CourseDailyMetrics records for the site for the matching day. The problem with this is that it will count users multiple times. For example, if a user is active in 2 different courses, then the user will get counted once for each course when added to the site cumulative active users instead of just once. This bug still needs to be addressed.

MAU may address this, meaning we may move the metric to MAU and remove the cumulative active user from SiteDailyMetrics, OR we may instead fix the cumulative in SiteDailyMetrics and use that for the stored MAU data. We need to put some thought into this 

cc @grozdanowski


johnbaldwin:
Thanks @melvinsoft !

PR #139, opened 2019-11-23
Fixes SiteDailyMetrics daily active users

johnbaldwin:
Now SiteDailyMetrics calcultes the daily active users for the site based
on the `modified` timestamp in `StudentModule` for the given site and
date

- Fixed a figures.tasks test: `test_populate_daily_mettics_error` was
raising an exception when it actually should have been just adding an
entry to the PipelineError model table
- Also did some Flake8 cleanup


tests/pipeline/test_site_daily_metrics.py
```diff
@@ -266,10 +288,6 @@ def test_load_for_today(self):
         for key, value in self.FIELD_VALUES.iteritems():
             assert getattr(sdm, key) == value, 'failed on key: "{}"'.format(key)
 
-
+    @pytest.mark.skip('Test stub')
     def test_no_course_overviews(self):
```
johnbaldwin: I'm leaving this test method here because we do want to test this, but it is a lower priority

PR #140, opened 2019-11-26
Adds course groups to Figures mocks

johnbaldwin:
- We need this to test working with the 'CohortMembership' model to
provide cohort data in Figures and DLReports

This is the first part of updating Figures test infrastructure to support DLReports


johnbaldwin:
Thanks @OmarIthawi !

PR #145, opened 2019-12-03
Test environment cleanup

johnbaldwin:
@OmarIthawi @melvinsoft This should be a quick one. Just some test code cleanup

Bump Python package versions for test packages for Hawthorn
- Also removed blank line from .coveragerc

Test helpers and test_metrics cleanup
- Removed tests.helpers.add_user_to_site because
  - Only tests.metrics.test_metrics used it
  - It is brittle code, relying on one org per site. We want to free
this constraint in our tests except exactly where needed to test
Site:Organization cardinality
  - Factories do the job cleaner

- Cleaned up tests.metrics.test_metrics to help improve single site/
multisite testing

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,4 +3,3 @@ omit =
     setup.py
     tests/*
     mocks/*
-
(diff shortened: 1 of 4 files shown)


johnbaldwin:
Thanks, that was FAST!

PR #147, opened 2019-12-04
Updated Figures MAU calculation to just use StudentModuel.modified

johnbaldwin:
- Also add new test

### DIFF
--- a/figures/metrics.py
+++ b/figures/metrics.py
@@ -231,11 +231,15 @@ def get_active_users_for_time_period(site, start_date, end_date, course_ids=None
 
     This is determined by finding the unique user ids for StudentModule records
     modified in a time period
+
+    We don't do this only because it raises timezone warnings
+        modified__range=(as_date(start_date), as_date(end_date)),
     """
     # Get list of learners for the site
+
     user_ids = figures.sites.get_user_ids_for_site(site)
     filter_args = dict(
-        created__gt=as_datetime(prev_day(start_date)),
+        modified__gt=as_datetime(prev_day(start_date)),
         modified__lt=as_datetime(next_day(end_date)),
         student_id__in=user_ids,
     )
--- a/tests/metrics/test_metrics.py
+++ b/tests/metrics/test_metrics.py
@@ -283,6 +283,33 @@ class TestSiteMetricsGettersStandalone(object):
                                                  end_date=self.data_end_date)
         assert count == len(student_module_sets)
 
+    def test_get_active_users_for_month(self):
+        date_before = datetime.date(2019, 8, 30)
+        dates_in = [
+            datetime.date(2019, 9, 1),
+            datetime.date(2019, 9, 15),
+            datetime.date(2019, 9, 30)
+        ]
+        start_date = dates_in[0]
+        end_date = dates_in[-1]
+        date_after = datetime.date(2019, 10, 1)
+        sm_out = [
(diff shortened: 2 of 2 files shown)


melvinsoft:
@johnbaldwin As I mentioned before, I'm not aware about the MAUs definition, but the code looks good.

abeals:
Before I levy my primary complaint relevant to the change, let me point out the Larger Issue hanging over this calculation. It's something we already discussed, so you know it, but I'm highlighting it because this PR doesn't fix it. (It's not an easy fix.)

The Larger Issue is that the value of `modified` changes over time, with no history. This means that recalculating MAU history is not possible. You have to calculate the value as soon as the month is over (and not a microsecond later!), and you have to store that value. You can't get it back later, and you can't look at StudentModule later to verify the accuracy of the stored number.

The fix involves keeping a separate, audit-able log of student activity events that can be used to recalculate MAUs in the case of aggregate data loss or pipeline failures. As noted above, this isn't an easy fix.

abeals:
Now, as for the change you're proposing.

The issue with using `modified` is one of timing of the monthly report execution. I had to draw a diagram to make sure I had it clear in my head.

IMG_8639

In this instance, the StudentModule is created in March, and the `modified` datetime is set to the same as `created` (I hope). Then there's a modified event later in March. So if you run the monthly report exactly at 0:00:00 on April 1st, and it takes zero microseconds to run the entire pipeline, then you've got accurate data.

Let's say that a student in a time zone where it's day time (or perhaps a night owl from the system's time zone) happens to take an action at 0:00:01 that results in a change to the `modified` field of the StudentModule. If the pipeline is run two milliseconds after midnight, or happens to hit that student a few seconds or minutes after midnight (after all, we're querying the database directly over time, not working from a perfect snapshot as of 0:00:00), then the student is no longer counted as an MAU for March.

I realize I'm painting a bit of a extreme example, but it's a problem with the approach we should acknowledge (in addition to the Larger Issue above).

One alternative would be to go only with `created`. That would allow for us to re-run the query any time in the future and get consistent, albeit inaccurate, data. In our example, the student would count in March, but not May. So let's toss that approach out, as it results in inaccurate data.

Another alternative would be to set up an OR query, doing something like:

```
from django.db.models import Q

or_query = Q(created__gt=as_datetime(prev_day(start_date)), modified__lt=as_datetime(next_day(end_date))) | Q(modified__gt=as_datetime(prev_day(start_date)), modified__lt=as_datetime(next_day(end_date)))

return StudentModule.objects.filter(or_query).values('student__id').distinct().count()
```

Again, this doesn't solve the Larger Issue, as a modified event in May would wipe out the April history if we ran the query later to back-fill. But it would give you more accuracy in running the MAU reports after the month close, so you wouldn't have to run the entire pipeline in zero milliseconds. (Which, by the way, if you figure out how to do, you should patent immediately.)

abeals:
In summary: It’s a series of tradeoffs we’re making here, choosing the "least worst" option until we've developed a solution to the Larger Issue. At that point, I expect that we'll be able to go back in history and recreate all MAU records consistently, like we did with MAUi's backfill capabilities. 

Until then, let's go with something that gets us as close to possible to the right answer.

**My preference, and recommendation:** Use the OR query. It's easy to understand, and it guards against modifications that occur between the end of month and query execution. It's _slightly_ more complicated because it's using a Q, but that's pretty standard Django ORM, nothing outlandish. Write a test for it, and make sure it's working as planned across the month boundary.

PR #154, opened 2019-12-23
Add a working spinner on API fetch waits

grozdanowski:
Currently Figures can struggle with fetching the data from a site with a large user base and course count. For example, in case of 3400 registered learners and total of 15000 course enrolments, we are looking at the API fetch pending state of around 25 seconds. This makes the app seem unresponsive (or the data not working at all). While we had a previous implementation of a spinner in that case, it did not work properly. This PR resolves this by tracking the Promises and their pending status.

Note: this needs a proper test on a staging server before clearing for production.


frontend/src/containers/loading-spinner/LoadingSpinner.js
```diff
@@ -1,32 +1,32 @@
...
-              />
-              <span>Loading your data...</span>
-            </div>
+  return (
+    <section className={styles['loading-spinner-root-container']}>
+      {promiseInProgress && (
+        <div className={styles['spinner-container']}>
+          <div className={styles['spinner-container__content']}>
+            <HashLoader
+              color={'#0090c1'}
```
johnbaldwin: Do we want to set the color code in a color file and name the variable so we don't have a magic number?

grozdanowski: Added it as a variable for more clarity.

PR #158, opened 2020-01-01
Ginkgo Backport

johnbaldwin:
This PR is to provide Ginkgo backward compatibility

## Highlights

* This should provide Ginkgo Backport functionality in a common branch to reduce maintenance costs

### Ginkgo test environment
* To enable shared code, the test suite uses an environment variable, `OPENEDX_RELEASE` with the value of "GINKGO" to use Ginkgo settings in the tests
* Added in `<figures project root>/mocks/ginkgo` mocks to mock Ginkgo edx-platform code
* Added new pytest config file, `pytest-ginkgo.ini` to specify the Ginkgo mocks directory
* Added Ginkgo environment to `tox.ini` to set 
* Added `ginkgo-env` environment setting file to set the Open edX version. This does `export OPENEDX_RELEASE=GINKGO`
* Added Ginkgo requirements file

### Production code
*  Updated serializers to support differences between Ginkgo and Hawthorn edx-platform code
* Updated migration to support differences in handling between Django 1.8 and 1.11
* Update Site Daily Metrics query filter to support the fact that Django 1.8 does not support `my_datetime_field__date` cast.
* Added `figures.compat` function "course_grade" to abstract the `CourseGradeFactory().[create|read]` function rename from Ginkgo to Hawthorn

### Devsite
* Updated devsite `settings.py` and `test_settings.py` for Ginkgo backward compatibility
* Updated the mock data seed generator code for Ginkgo backward compatibility

## What is left?
* Need to retest with Ginkgo Vagrant virtualenv
* Need to test with a Ginkgo enterprise staging server
* Need to test on Tahoe staging
* Need to test in a Hawthorn enterprise staging server


johnbaldwin:
test_compat breaking now locally with tox. Working on a fix

johnbaldwin:
Simple one-liner fixed that. Need to add `@patch('openedx.core.release.RELEASE_LINE', 'hawthorn')` for the failing test

johnbaldwin:
@thraxil Any idea on why `codecov/patch` seems to get stuck with "Expected — Waiting for status to be reported"?

OmarIthawi:
@johnbaldwin it's probably configured in the [`figures` branches settings](https://github.com/appsembler/figures/settings/branches). I thin you can safely remove it from there as a requirement since `codecov/project` seems to be working okay.

devsite/devsite/settings.py
```diff
@@ -10,10 +10,16 @@
 import sys
 
 from figures.settings.lms_production import (
-    update_celerybeat_schedule,
     update_webpack_loader,
+    update_celerybeat_schedule,
 )
 
+OPENEDX_RELEASE = os.environ.get('OPENEDX_RELEASE', 'HAWTHORN').upper()
```
OmarIthawi: Probably we should have had something like this:

```
settings/common.py
settings/hawthorn.py
settings/ginkgo.py
```

That would make it more readable.

ginkgo-env
```diff
@@ -0,0 +1 @@
+export OPENEDX_RELEASE='GINKGO'
```
OmarIthawi:
```suggestion
```

Since tox already have that variable.

pytest-ginkgo.ini
```diff
@@ -0,0 +1,10 @@
+# This file supports PyTest testing Figures against Ginkgo mocks
+
+[pytest]
```
OmarIthawi: `tox` should be used for tests multiplexing, I think it should be used instead of doing custom work like this.

devsite/devsite/settings.py
johnbaldwin: For a newcomer to the code, it would be more readable, but then we would need six environment files and then we'd have to maintain them. Then we would feel the pressure to abstract the common settings, then it gets complicated. I think having only two settings files, one for devsite and one for tests is a fair tradeoff. What could help is to make it clear how settings work in the module docstring

johnbaldwin: We would also need settings files for each of the `test_<platform>.py` setting files. I'm also not entirely convinced shared base settings files is a good pattern. I've seen it lead to unnecessary complexity and timing issues

ginkgo-env
johnbaldwin: I use it for devsite ginkgo (and I ought to add instructions for running devsite in "Ginkgo mode".  When I update the `Makefile` to act as a full wrapper around devsite for Ginkgo, then I'll delete this file

pytest-ginkgo.ini
johnbaldwin: @OmarIthawi I looked at using env vars for `pytest.ini` The only project I found is no longer maintained. Do you know of another way besides env vars to set the `mocks/<platform>` directory to be a variable based on the tox environment for pytest?. For Ginkgo, we need to set:

```
[pytest]
...
python_paths = devsite mocks/ginkgo
```

and for Hawthorn, we need to set:

```
[pytest]
...
python_paths = devsite mocks/hawthorn
```

OmarIthawi: @johnbaldwin I'm not sure what the question is, could you please explain?

In `tox.ini` files it's possible to manipulate the python path:

```ini
[testenv]
commman = pytest

[testenv:py27-ginkgo]
setenv =
    PYTHONPATH = devsite mocks/ginkgo

[testenv:py27-hawthorn]
setenv =
    PYTHONPATH = devsite mocks/hawthorn
```

Is that what you're looking for?

OmarIthawi:
@johnbaldwin could you please keep the original review and avoid dismissing it? We are not yet ready for such strict review practices due to lack of resources.

pytest-ginkgo.ini
OmarIthawi: nvm, that was not needed. Thanks John.

johnbaldwin:
@grozdanowski Is Bryan's comment something you can address?

johnbaldwin:
@bryanlandia I'm not entirely clear on your comment. Is this strictly on the front end or are we also looking at version issues with the Python package 'django-webpack-loader'

cc @grozdanowski Can you weigh in here?

grozdanowski:
> @johnbaldwin Ginkgo uses an earlier version (`^2.2.1` or any 2.x series) of webpack instead of 4.x+ so some of the dependencies need to be looked at.
> 
> The only actual failure I found running in Ginkgo devstack React app was that `npm start` fails with
> 
> ```
> npm WARN css-loader@3.4.1 requires a peer of webpack@^4.0.0 || ^5.0.0 but none was installed.
> ```
> 
> css-loader from release 1.0 onward requires webpack 4.x+. I was able to start the development server using the last pre-1.0 release of css-loader.
> 
> ```
> diff --git a/frontend/package.json b/frontend/package.json
> index 4dee131..6463f3c 100644
> --- a/frontend/package.json
> +++ b/frontend/package.json
> @@ -121,7 +121,7 @@
>      "extends": "react-app"
>    },
>    "devDependencies": {
> -    "css-loader": "^3.2.0",
> +    "css-loader": "<1.0",
>      "webpack-dev-server": "2.9.4"
>    }
>  }
> ```

@bryanlandia what do you mean by "Ginkgo uses"?

johnbaldwin:
Merged master into this branch. Now investigating what the problem is with `webpack-stats.json`

johnbaldwin:
@bryanlandia Is the front end compilation still an issue for you?

bryanlandia:
Sorry @johnbaldwin I haven't gotten back to test the build with yarn but will try today or tomorrow.   I'll release my change request

johnbaldwin:
Thanks @bryanlandia ! Please let Matej and me know if you run into any problems building the front end assets on your dev machine with yarn

Makefile
```diff
@@ -50,6 +50,11 @@ pylint:  ## lint
 coverage:  ## Run coverage, without the built-in virtualenv
 	@. ve/bin/activate; coverage run --source figures -m py.test; coverage report -m
 
+ginkgo.pytest:  ## Run Pytest for the Ginkgo environment
+	OPENEDX_RELEASE=GINKGO pytest -c pytest-ginkgo.ini
+
+ginkgo.tox:  ## Run tox just for the Ginkgo environment
+	tox -e py27-ginkgo
```
bryanlandia: I can't run any of the `make` commands, failing on 

```
Makefile:5: .env*: No such file or directory
```
Are those dotenv files generated or specified somehow else that I'm not seeing, or are those particular to your dev environment @johnbaldwin ?  Is there any requied configuration that should be stored in them?  Shouldn't this not fail if missing?

devsite/requirements/ginkgo.txt
```diff
@@ -0,0 +1,70 @@
...
+##
+
+Django==1.8.18
+django-extensions==1.5.9
+# ./edx/base.txt:git+https://github.com/edx/django-rest-framework.git@3c72cb5ee5baebc4328947371195eae2077197b0#egg=djangorestframework==3.2.3
+djangorestframework==3.2.3
+django-countries==4.0
+django-filter==0.11.0
+django-webpack-loader==0.5.0
+# appsembler/gingko/master users 0.4.1
```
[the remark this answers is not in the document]
johnbaldwin: @bryanlandia I think this is a leftover from when Tahoe was on Ginkgo. I changed it to `0.4.1` and pushed the update to this PR

Makefile
bryanlandia: Just need a look at this @johnbaldwin and I'll approve :D

johnbaldwin: @bryanlandia I've been building the makefile as a helper, so I think you are the first other person to run it. There should be a default .env file, I'll add as a different PR. To note, TravisCI does not use the makefile.

 Please ignore this for this PR and just run this:

`tox -e py27-ginkg`

johnbaldwin: But if you are curious, the `.env` file looks something like this:

```
# This file provides default environemnt settings
# override by creating a '.env.<some-label>' file for the specific settings
# you wish to override

# Set the edx-platform named release
EDX_PLATFORM_RELEASE=hawthorn
```

bryanlandia: @johnbaldwin OK, sounds fine.  I did run that `tox` command.

PR #159, opened 2020-01-05
Avoid GeneratedCertificate import errors

OmarIthawi:
Using explicit RELEASE_LINE check to avoid the issue.

I don't have a devstack to verify it, but I think it's possible that `lms.djangoapps.certificates.models import GeneratedCertificate` is succeeding on Ginkgo and that would confuse the Django database migration system.

Doing the explicit check ensures that we don't have this issue.

### DIFF
--- a/figures/compat.py
+++ b/figures/compat.py
@@ -26,12 +26,11 @@ except ImportError:
         # try the pre-ginkgo path
         from lms.djangoapps.grades.new.course_grade import CourseGradeFactory    # noqa: F401
 
-try:
-    # First try to import for the path as of hawthorn
-    from lms.djangoapps.certificates.models import GeneratedCertificate
-except ImportError:
-    # try the pre-hawthorn path
-    from certificates.models import GeneratedCertificate    # noqa: F401
+
+if RELEASE_LINE == 'ginkgo':
+    from certificates.models import GeneratedCertificate  # noqa: F401
+else:
+    from lms.djangoapps.certificates.models import GeneratedCertificate  # noqa: F401
 
 
 def course_grade(learner, course):
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -7,7 +7,7 @@ NOTE
 ====
 
 Pytest monkeypatch does not support testing imports as it does not undo updating
-entries in sys.modules. Therefore, we need to use mock.patch context manager.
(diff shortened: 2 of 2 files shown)


figures/compat.py
```diff
@@ -26,12 +26,13 @@
         # try the pre-ginkgo path
         from lms.djangoapps.grades.new.course_grade import CourseGradeFactory    # noqa: F401
 
-try:
```
johnbaldwin: @OmarIthawi  Why not just do this? (and not add the extra function)

```
if RELEASE_LINE == 'ginkgo':
    from certificates.models import GeneratedCertificate
else:
    from lms.djangoapps.certificates.models import GeneratedCertificate
```

Also I seem to remember a PR review comment you made to me that I should avoid using getter methods when we can just import the object directly in the flavor of 'Six'

OmarIthawi: Sure thing @johnbaldwin! Done.

tests/test_compat.py
```diff
@@ -16,77 +16,89 @@
...
 
-import mock
+from mock import Mock, patch
 import pytest
-import types
+import six
+from types import ModuleType
 
 
+def patch_module(module_path, extra_properties=None):
```
johnbaldwin: Awesome @OmarIthawi Thanks!

johnbaldwin:
@OmarIthawi First, thanks for the update! I do notice that the `py27-ginkgo` tox environment is not being run by Travis. This environment is failing for me when I run `tox -e py27-ginkgo` locally. So I'm looking into it. It is failing with `ImportError: No module named course_groups

OmarIthawi:
Thanks for pointing out @johnbaldwin.

I checked and found that the same error occurs in the `john/ginkgo-backport`. I think it's out of the scope of this pull request.

johnbaldwin:
@OmarIthawi I'm planning on fixing it.I don't have a preference for which branch the fix goes in. Do you?

OmarIthawi:
@johnbaldwin my preference is that it would be fixed in the `john/ginkgo-backport`, following the same usual workflow. Preferably after merging this pull request to avoid having two persons working on the same local branch.

johnbaldwin:
@OmarIthawi Sounds good

PR #165, opened 2020-01-14
Enable `codecov/patch` again

OmarIthawi:
This should solve the issue [**RED-657:** Codecov/patch gets stuck with the message "Expected — Waiting for status to be reported" ](https://appsembler.atlassian.net/browse/RED-657).

### A bit of a background
During the Ginkgo backport John have asked to disable the `patch` because it was broken without an immediate solution so Anders disabled `codecov/patch` in #150, this PR should enable the `codecov/patch` again because we wanted it again now.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,0 @@
-coverage:
-  status:
-    patch: false
-    project:
-      default:
-        target: 85%
-        threshold: 10%


johnbaldwin:
@OmarIthawi Thanks. For clarification, `codecov/patch` wasn't noisy. It just wasn't passing due to code coverage dropping for the ginkgo backport. The code coverage dropping _appears_ to be because codecov was reporting Ginkgo and Hawthorn coverage independently. I investigated a little bit on that, but didn't find anything conclusive so stopped spending time on it.

I just now re-enabled `codecov/patch` in the branch rules settings and re-started the travis build. Let's see how that works

OmarIthawi:
@johnbaldwin it works now.

PR #169, opened 2020-01-27
Enable pylint with all messages disabled by default

OmarIthawi:
[**RED-668:** Pylint Figures](https://appsembler.atlassian.net/browse/RED-668).

I'll issue a pull request or two to incrementally enable checks and resolve them once this has landed one master.

This PR enables the [`edx_lint`](https://github.com/edx/edx-lint) package which is nice to manage a huge [`pylintrc`](https://github.com/appsembler/course-access-groups/blob/master/pylintrc) file and skim it into a few lines of [`pylintrc_tweaks`](https://github.com/appsembler/course-access-groups/blob/master/pylintrc_tweaks).


johnbaldwin:
@OmarIthawi 

A. We're using edX lint checking instead of pylint alone. Is edX lint specific to running on edX code or plugins or can it be used by any Python project? What are the advantages?

B. Is the pylintrc_tweaks file automatically read by `edx_lint`? I assume yes, just wanted to make sure I'm not missing something

C. When would we need to regenerate the pylintrc file? Would we need to regenerate the pylintrc_tweaks file or now that it is created, we're free to modify it without concern that it will be regenerated?

OmarIthawi:
@johnbaldwin 

> A. We're using edX lint checking instead of pylint alone. Is edX lint specific to running on edX code or plugins or can it be used by any Python project? What are the advantages?

We're still using `pylint` for the actual check. `edx_lint` is a tool to manage `pylintrc` and keep it consistent with the other upstream edX code. As you'd guess I like to stay close to the upstream standards because of a number of benefits including:

 - It saves time to re-think what a good pylint rule look like, I'd rather copy some template and focus on the application code instead of linters and rules.
 - It makes it easier for potential community contributors to push code to Figures.
 - I don't have to follow two different standards when working on internal and upstream code.

> B. Is the pylintrc_tweaks file automatically read by `edx_lint`? I assume yes, just wanted to make sure I'm not missing something

Not really. `edx_lint` is a just a tool to generate `pylintrc` files. I'll refer to the [edx_lint docs](https://github.com/edx/edx-lint#using-edx_lint).
 
> C. When would we need to regenerate the pylintrc file? Would we need to regenerate the pylintrc_tweaks file or now that it is created, we're free to modify it without concern that it will be regenerated?

When `pylintrc_tweaks` is changed, we need to run `$ tox -e wirte_edx_lint`. If we forgot the CI will tell us because I added a rule to avoid such issue of having `pylintrc` and `pylintrc_tweaks` out of sync.

johnbaldwin:
@OmarIthawi We need to merge master back to this PR. Do you want me to do that?

And thanks for the answers!

PR #171, opened 2020-01-29
Bump Figures version to 0.3.4 and update main README

johnbaldwin:
Updated the main project README to provide some information on release 0.3.4. The copy in the readme was woefully out of date, so I just removed the content so I would not be spending time on it now. I have an open card to update docs.

Bumped figures version to 0.3.4 in `setup.py`


johnbaldwin:
Thanks @OmarIthawi !

PR #173, opened 2020-02-06
API update - Adds SiteMonthlyMetrics viewset

johnbaldwin:
SiteMonthlyMetrics viewset
* Adds new endpoints to retrieve metrics currently in 'GeneralSiteMetricsViewSet'
* endpoints have underscore in url for the custom actions. Deferring
fix. Should create a bug card for this

Other fixes and updates in this commit:

* Fixed figures.metrics naive datetime objects
* Updated figures.metrics to use the User model directly for registered
users (instead of getting from SiteDailyMetrics)

### DIFF
--- a/figures/helpers.py
+++ b/figures/helpers.py
@@ -111,7 +111,8 @@ def prev_day(val):
 
 
 def days_in_month(month_for):
-    return calendar.monthrange(month_for.year, month_for.month)[1]
+    _, num_days_in_month = calendar.monthrange(month_for.year, month_for.month)
+    return num_days_in_month
 
 
 # TODO: Consider changing name to 'months_back_iterator' or similar
--- a/figures/helpers.py
+++ b/figures/helpers.py
@@ -110,6 +110,10 @@ def prev_day(val):
     return days_from(val, -1)
 
 
+def days_in_month(month_for):
+    return calendar.monthrange(month_for.year, month_for.month)[1]
+
+
 # TODO: Consider changing name to 'months_back_iterator' or similar
 # TODO: implement or removed include_month_for
 def previous_months_iterator(month_for, months_back, include_month_for=False):
@@ -128,5 +132,5 @@ def previous_months_iterator(month_for, months_back, include_month_for=False):
         start_month = month_for - relativedelta(months=months_back)
 
     for dt in rrule(freq=MONTHLY, dtstart=start_month, until=month_for):
-        last_day_of_month = calendar.monthrange(dt.year, dt.month)[1]
+        last_day_of_month = days_in_month(month_for=dt)
         yield (dt.year, dt.month, last_day_of_month)
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,15 +1,60 @@
-'''Test Figures Django management commands
+"""Test Figures Django management commands
+"""
 
-'''
-
+import mock
+import pytest
 from django.core.management import call_command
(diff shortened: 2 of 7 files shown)


johnbaldwin:
Added test for MAU option in the `populate_figures_metrics` Django management command to get the 'codecov/patch' check to pass

figures/helpers.py
```diff
@@ -110,6 +110,10 @@ def prev_day(val):
     return days_from(val, -1)
 
 
+def days_in_month(month_for):
+    return calendar.monthrange(month_for.year, month_for.month)[1]
```
OmarIthawi: No biggie, better than having the magical constant `1`.

```suggestion
    """
    Calculate days in a certain month.
    """
    _from_day, to_day = calendar.monthrange(month_for.year, month_for.month)[1]
    return to_day
```

tests/test_commands.py
```diff
@@ -1,15 +1,60 @@
...
+def test_mau_no_delay_with_mock(transactional_db):
+    """
+    This would be the preferred test to the above. Acccording to the documentation,
+    this should work, but the mock never gets called, so something is not right
+
+    https://docs.python.org/dev/library/unittest.mock.html#where-to-patch
+    """
+    path = 'figures.management.commands.populate_figures_metrics.populate_all_mau'
+    with mock.patch(path) as mock_populate_all_mau:
+        call_command('populate_figures_metrics', '--no-delay')
```
OmarIthawi: Thanks for documenting that it's not working. I tested it, and looks like the `--mau` argument is missing. This should fix it:

```suggestion
        call_command('populate_figures_metrics', '--no-delay', '--mau')
```

tests/test_commands.py
```diff
@@ -1,15 +1,60 @@
...
     def test_command_output(self):
         out = StringIO()
         call_command('populate_figures_metrics', '--no-delay', stdout=out)
 
         self.assertEqual('', out.getvalue())
-        #self.assertIn('Expected output', out.getvalue())
\ No newline at end of file
+
+
+def test_mau_no_delay(monkeypatch):
```
OmarIthawi:
```suggestion
```

Thanks for documenting this, please see below. I think I've fixed the original function, so this one isn't needed.

tests/test_commands.py
```diff
@@ -1,15 +1,60 @@
...
+        call_list.append(True)
+
+    path = 'figures.management.commands.populate_figures_metrics.populate_all_mau'
+    monkeypatch.setattr(path, mock_populate_all_mau)
+    call_command('populate_figures_metrics', '--no-delay', '--mau')
+    assert len(call_list) == 1
+
+
+@pytest.mark.skip('This test does not work')
+def test_mau_no_delay_with_mock(transactional_db):
```
OmarIthawi:
```suggestion
def test_mau_no_delay(transactional_db):
```

figures/helpers.py
johnbaldwin: @OmarIthawi Sure, we'll do that, but we need to drop the `[1]` from the call you suggest, and the first value is the day of the week of the first day of the month. This is the new function:

```
def days_in_month(month_for):
    _, num_days_in_month = calendar.monthrange(month_for.year, month_for.month)
    return num_days_in_month
```

tests/test_commands.py
johnbaldwin: Thanks @OmarIthawi !

PR #174, opened 2020-02-06
Fix MAU view test for multisite mode

johnbaldwin:
To test multisite mode, figures multisite mode need to be set to True

### DIFF
--- a/tests/views/test_mau_views.py
+++ b/tests/views/test_mau_views.py
@@ -30,6 +30,8 @@ class TestSiteMauLiveMetricsViewSet(BaseViewTest):
 
     @pytest.fixture(autouse=True)
     def setup(self, db, settings):
+        if organizations_support_sites():
+            settings.FEATURES['FIGURES_IS_MULTISITE'] = True
         super(TestSiteMauLiveMetricsViewSet, self).setup(db)
 
     def test_list(self, monkeypatch, sm_test_data):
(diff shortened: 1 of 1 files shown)


tests/views/test_mau_views.py
```diff
@@ -30,6 +30,8 @@ class TestSiteMauLiveMetricsViewSet(BaseViewTest):
 
     @pytest.fixture(autouse=True)
     def setup(self, db, settings):
+        if organizations_support_sites():
+            settings.FEATURES['FIGURES_IS_MULTISITE'] = True
```
OmarIthawi: Will this clean up after the test is done? I haven't used [`pytest` settings myself](https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) to know.

OmarIthawi:
```suggestion
        settings.FEATURES['FIGURES_IS_MULTISITE'] = organizations_support_sites()
```

Although this is practically the same, in unit tests `if` statements are not recommended.

johnbaldwin: Good suggestion, thanks @OmarIthawi !

johnbaldwin: Good question, I'll find out. For our case it doesn't matter because the 'orgs support sites' is global for the environment since the `edx-organizations` package choice determines what the result of `organizations_support_sites` function call will be

johnbaldwin:
Getting a yarn error on the lint check: https://travis-ci.org/appsembler/figures/jobs/647422503#L522

Since the tox lint environment is to inspect the Python code, there should be no need to run yarn for this environment

PR #175, opened 2020-02-06
Fix MAU view test for multisite mode

OmarIthawi:
To test multisite mode, figures multisite mode need to be set to True

Closes #174


OmarIthawi:
Checking the [`pytest` docs](https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) I'm not sure if this is necessary to have.

PR #177, opened 2020-02-11
Makefile 'make coverage' fix to use default python env

johnbaldwin:
Removing running coverage in the bundled virtualenv because I still need to work on it and run coverage for both Tahoe and community vents


johnbaldwin:
Thanks Omar!

PR #179, opened 2020-02-11
Optimise autocomplete search

grozdanowski:
## This PR addresses the following issues:
- slow initial load time of the app because complete User and Course indexes are loaded into Redux
- certain users are not accessible because of pagination in those APIs

## How does it address these issues:
- we remove those indexes from Redux and initial render-blocking fetches
- we dynamically fetch from respective APIs upon user input into these autocompletes
- BUT we don't just fetch as the user types. We do a timeout, so the fetch is done ONLY after the user has not typed anything into the input for 1000 milliseconds. This allows us to optimise the number of API calls.

## How it now looks:

autocomplete-deferred


grozdanowski:
@OmarIthawi @johnbaldwin I have rebased this off Master to include additional API options that @melvinsoft did and that were merged in today. Autocomplete searches now pull the data dynamically and it works nicely and the load is reduced.

johnbaldwin:
I merged master into this branch (And now only 8 changed files instead of 43)

frontend/src/components/inputs/AutoCompleteUserSelect.js
```diff
@@ -46,11 +40,25 @@ class AutoCompleteUserSelect extends Component {
...
+
     this.setState({
       value: newValue
     });
+
+    this.timer = setTimeout(this.doSearch, WAIT_INTERVAL);
   };
 
+  doSearch = () => {
+    const requestUrl = apiConfig.learnersGeneral + '?search=' + this.state.value;
```
OmarIthawi:
```suggestion
    const requestUrl = apiConfig.learnersGeneral + '?search=' + encodeURIComponent(this.state.value);
```

frontend/src/components/inputs/AutoCompleteCourseSelect.js
```diff
@@ -46,11 +40,25 @@ class AutoCompleteCourseSelect extends Component {
...
+
     this.setState({
       value: newValue
     });
+
+    this.timer = setTimeout(this.doSearch, WAIT_INTERVAL);
   };
 
+  doSearch = () => {
+    const requestUrl = apiConfig.coursesGeneral + '?search=' + this.state.value;
```
OmarIthawi:
```suggestion
    const requestUrl = apiConfig.coursesGeneral + '?search=' + encodeURIComponent(this.state.value);
```

johnbaldwin: Cool, @OmarIthawi I made a note of this for when I get back into JavaScript programming again

frontend/src/components/inputs/AutoCompleteCourseSelect.js
```diff
@@ -1,23 +1,16 @@
...
 
-var coursesList = [
-];
-
-const getSuggestions = value => {
-  const inputValue = value.trim().toLowerCase();
-  const inputLength = inputValue.length;
-  return inputLength === coursesList ? [] : coursesList.filter(course => ((course.courseName.toLowerCase().slice(0, inputLength) === inputValue) || (course.courseNumber.toLowerCase().slice(0, inputLength) === inputValue))).toArray();
-};
+const WAIT_INTERVAL = 1000;
```
johnbaldwin: @grozdanowski How helpful would being able to pull configuration data from the server? I see this as an example where we might want to tune what the wait interval is without having to deploy updated code

grozdanowski: @johnbaldwin we can pass this as an env variable.  I think it's outside the scope for this PR, but we could do quite a lot of settings of the app that way, making it more flexible.

PR #185, opened 2020-02-17
Refactor figures.permissions

johnbaldwin:
This update simplifies Figures multisite organization admin user check.
The motivation is from Omar Al-Ithawi's PR: https://github.com/appsembler/course-access-groups/pull/12 for the `is_organization_staff` function in `acl_backends.py`

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -1,14 +1,10 @@
-'''Test the permission classes we define in Figures
+"""Test the permission classes we define in Figures
+"""
 
-'''
-
-import mock
 import pytest
 
 from django.contrib.auth import get_user_model
-
 import django.contrib.sites.shortcuts
-
 from rest_framework.test import APIRequestFactory
 
 import figures.permissions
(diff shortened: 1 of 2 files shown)


johnbaldwin:
`codecove/patch` appears to have dropped coverage because coverage is only running for the community environment and not also for multisite. So we probably want to add another coverage environment that runs the multisite requirements specified

johnbaldwin:
`coverage/patch` is skipping over the code when multisite code, but this code *is* being tested with the `py27-hawthorn_multisite` tox environment. Looks like coverage is not tracking coverage for multisite tests

tests/test_permissions.py
```diff
@@ -130,20 +124,33 @@ def test_site(request):
...
         monkeypatch.setattr(django.contrib.sites.shortcuts, 'get_current_site', test_site)
         settings.FEATURES['FIGURES_IS_MULTISITE'] = True
         assert figures.helpers.is_multisite()
         org2 = OrganizationFactory(sites=[self.site])
-        UserOrganizationMappingFactory(user=request.user, organization=org2),
-        with pytest.raises(figures.permissions.MultipleOrgsPerUserNotSupported):
-            figures.permissions.IsSiteAdminUser().has_permission(request, None)
+        UserOrganizationMappingFactory(user=request.user, organization=org2)
+        permission = figures.permissions.IsSiteAdminUser().has_permission(request, None)
+        assert permission is allow, 'User "{username}" should have access'.format(
```
OmarIthawi:
```suggestion
        assert permission == allow, 'User "{username}" should have access'.format(
```

I don't think `is` should be used for primitives. There's a reason, but I forgot why :sweat_smile:

johnbaldwin: I'll change it, thanks @OmarIthawi !

PR #187, opened 2020-02-18
Added Admin interface to Course MAU Metrics model and updated admin test

johnbaldwin:


figures/admin.py
```diff
@@ -85,3 +85,14 @@ class PipelineErrorAdmin(admin.ModelAdmin):
...
+
+
+@admin.register(figures.models.CourseMauMetrics)
+class CourseMauMetricsAdmin(admin.ModelAdmin):
+    """Defines the admin interface for the CourseMauMetrics model
+    """
+    list_display = ('id', 'date_for', 'site', 'course_id', 'mau')
+    list_filter = (
+        ('site', RelatedOnlyDropdownFilter),
+        ('course_id', AllValuesDropdownFilter),
```
OmarIthawi: That'll be a bit slow, no?

johnbaldwin: It is slow when there are are a lot of records, but at least it is a starting point

johnbaldwin:
Thanks @OmarIthawi !

PR #192, opened 2020-02-22
Check for and fix 5 pylint warnings (round 3)

OmarIthawi:
Fixing the following errors:

 - `attribute-defined-outside-init`: Initialize the attribute with `None` e.g. `site = None`
 - `inconsistent-return-statements`: Make it consistent by `return None`
 - `len-as-condition`: Simply removing `len()` fixes the issue
 - `redefined-builtin`: Renaming a couple of builtins like `id -> user_id` and `format -> fmt`
 - `unused-argument`: Remove the argument, prefix it with underscore e.g. `_kwargs` or use a pylint disable flag

### DIFF
--- a/figures/metrics.py
+++ b/figures/metrics.py
@@ -53,10 +53,10 @@ import figures.sites
 #
 
 
-def period_str(month_tuple, format='%Y/%m'):
+def period_str(month_tuple, fmt='%Y/%m'):
     """Returns display date for the given month tuple containing year, month, day
     """
-    return datetime.date(*month_tuple).strftime(format)
+    return datetime.date(*month_tuple).strftime(fmt)
 
 
 #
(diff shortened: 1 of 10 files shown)


figures/metrics.py
```diff
@@ -91,7 +91,7 @@ class LearnerCourseGrades(object):
     TODO: Make convenience method to instantiate from a GeneratedCertificate
     """
 
-    def __init__(self, user_id, course_id, **kwargs):
+    def __init__(self, user_id, course_id, **_kwargs):
```
melvinsoft: I'm learning here, why do we need the underscore here?

OmarIthawi: @melvinsoft `**kwargs` is unused. Pylint has a bug and cannot supress the warning: `unused-argument` until Pylint v2. The solution is to use another convention for ignored variables which is to use `_` prefix to denote an unused variable.

PR #196, opened 2020-03-12
Spike to test StudentModule save notification receiver

johnbaldwin:
This is spike code to run on Tahoe staging and have folks do some course activity. This will validate that the django model `post_save` signal notifies our receiver. This spike just dumps data to the default log with log lines starting with `StudentModule.saved` for easy grep'ing. 

* https://docs.djangoproject.com/en/1.11/ref/signals/#post-save

# PyLint failing

Pylint is failing in `figures.app` for the `import figures.signals` line even though I've added lint error suppression there

```
************* Module figures.apps
659W: 62, 8: Unused variable 'figures.signals' (unused-variable)
660I: 62, 0: Useless suppression of 'unused-import' (useless-suppression)
```


johnbaldwin:
@OmarIthawi Any idea why pylint is still complaining about the import in `figures.apps` even though the error suppression comment line is appended to the import? I also tried `# noqa: F401` but that does not suppress the error either

figures/apps.py
```diff
@@ -57,3 +57,6 @@ class FiguresConfig(AppConfig):
                 }
             },
         }
+
+    def ready(self):
+        import figures.signals  # noqa pylint: disable=unused-import
```
OmarIthawi: Pylint is complaining about `unused-variable` instead of `unused-import`.

```suggestion
        import figures.signals  # noqa pylint: disable=unused-variable
```

thraxil:
Is this still being worked on?

PR #199, opened 2020-03-17
Optimise api calls

grozdanowski:
This PR optimises the load times of the Figures app by:
- removing the call to course details API on Dashboard loading (that became obsolete)
- removing the initial call to pre-populate the dropdowns for Course and User header autocompletes. That is not essential (low UX impact estimated) and shaves a lot off initial load times.

### DIFF
--- a/frontend/src/components/inputs/AutoCompleteCourseSelect.js
+++ b/frontend/src/components/inputs/AutoCompleteCourseSelect.js
@@ -83,12 +83,6 @@ class AutoCompleteCourseSelect extends Component {
 
   componentWillMount() {
     this.timer = null;
-    fetch((apiConfig.coursesGeneral), { credentials: "same-origin" })
-      .then(response => response.json())
-      .then(json => this.setState({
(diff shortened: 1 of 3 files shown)


frontend/src/components/inputs/AutoCompleteCourseSelect.js
```diff
@@ -83,12 +83,6 @@ class AutoCompleteCourseSelect extends Component {
 
   componentWillMount() {
     this.timer = null;
-    fetch((apiConfig.coursesGeneral), { credentials: "same-origin" })
-      .then(response => response.json())
-      .then(json => this.setState({
-        suggestions: json['results'],
```
OmarIthawi: Don't we need to remove the references to it as well? or that's already removed.

grozdanowski: It is just the initial API call to pre-populate the dropdown. It's nice to have, but quite non-essential.

PR #202, opened 2020-04-03
Refactor frontend site/course data API fetching

grozdanowski:
This PR addresses the performance issues, as well as the bloated Redux store the app had so far. Most notably, we were previously using the general site metrics API endpoint which had a terrible impact on the performance and load times.

We are now taking a different approach, with new API endpoints on the backend that allow us to fetch the data in a more granular way (and super duper quicker). To complement that, the frontend was reworked a bit:

- the BaseStatCard can now obtain `valueHistory` and `mainValue` parameters (props) in multiple ways:
-- static values (passed strings, numbers, or values passed from Redux for instance)
-- an API fetch abstraction function that allows it to fetch the data dynamically (without blocking the rendering of rest of the app)
-- a combination where for instance it will load the value statically (for instance from Redux so the value is immediately available or not reloaded unnecessarily each time) and fetch its history asynchronously

This PR implements these changes into the Dashboard view, MAU history view, single Course view.

The number of API calls has actually increased, but with a lot of them not being render blocking and each of them fetching _significantly_ less data (and getting a _way_ faster response from the backend), it gives us an improvement to performance of the app.

Such performance. Much wow.


OmarIthawi:
Reviewing that in a sec!

PR #210, opened 2020-04-26
Added Django Debug Toolbar to the devsite

johnbaldwin:
This is mostly helpful to help debug slow queries and could be helpful for other issues too.

Please try it out in your dev environment

https://github.com/jazzband/django-debug-toolbar

There is one pretty annoying thing that I'm still investigating by digging into the debug toolbar source code. The SQL history just keeps growing and growing and there is not a clear way to reset/clear the SQL history.


johnbaldwin:
Thanks for the review, Matej and Omar!

PR #214, opened 2020-05-13
Enrollment Metrics Rework

johnbaldwin:
Added new `figures.pipeline.enrollment_metrics` module to conditionally retrieve progress data from the platform depending on the states of `StudentModule` and `LearnerCourseGradeMetrics` models for the given `CourseEnrollment`

Running `CourseGradeFactory().read(..)` is costly and needed to retrieve dynamic grades when persistent grades is not enabled or supported (as is the case with Ginkgo). This PR addresses this

The existing `get_average_progress` method is renamed to `get_average_progress_deprecated` instead of just deleting it. This is so we have the code available in production to support troubleshooting and validation

The individual commits, commit messages and inline documentation should explain details for this PR


johnbaldwin:
I need to look into this. Pylint seems to think that the queryset is a list

https://travis-ci.org/github/appsembler/figures/jobs/686429867#L665

The code it complains about is this:

```
    most_recent_lcgm = LearnerCourseGradeMetrics.objects.filter(
        user=course_enrollment.user,
        course_id=str(course_enrollment.course_id)).order_by('date_for').last()

File: figures/pipeline/enrollment_metrics.py [hunk 1]
```diff
@@ -0,0 +1,272 @@
...
+
+Calls are reduced by checking if the learner has a `StudentModule` record
+modified after the most recent `LearnerCourseGradeMetrics` record `date_for`
+field
+
+
+# Logic
+
+```
+for each learner+course
```
OmarIthawi, line 16: Great section! Would you mind adding references in the logic to which function I should take a look at e.g.

```suggestion
for each learner+course  # see `main` function
```

figures/pipeline/enrollment_metrics.py
```diff
@@ -0,0 +1,272 @@
...
+    get newest lgcm record
+    get newest sm record
+
+    if not lcgm and not sm
```
OmarIthawi, line 20:
```suggestion
    if not lcgm and not sm  # see `func_123`
```

figures/pipeline/enrollment_metrics.py
johnbaldwin: Will do!

johnbaldwin: Actually, doing this for each line will add more noise than value, making the logic harder to read.  I'll at least add in references about the core iteration and update check

johnbaldwin:
Fixed the pylint failure. Looks like an issue with pylint-django: https://github.com/PyCQA/pylint-django/issues/165

PR #224, opened 2020-06-12
Added db_index=True to LearnerCourseGradeMetrics model

johnbaldwin:
Query performance was very slow for this model. Added indexing to the
fields which we use for filtering

### DIFF
--- a/figures/models.py
+++ b/figures/models.py
@@ -206,11 +206,11 @@ class LearnerCourseGradeMetrics(TimeStampedModel):
     actually needed and edx-platform uses FloatField in its grades models
 
     """
-    site = models.ForeignKey(Site)
-    date_for = models.DateField()
+    site = models.ForeignKey(Site, db_index=True)
+    date_for = models.DateField(db_index=True)
     # TODO: We should require the user
-    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
-    course_id = models.CharField(max_length=255, blank=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True, db_index=True)
+    course_id = models.CharField(max_length=255, blank=True, db_index=True)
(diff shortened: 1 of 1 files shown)


figures/models.py
```diff
@@ -206,11 +206,11 @@ class LearnerCourseGradeMetrics(TimeStampedModel):
     actually needed and edx-platform uses FloatField in its grades models
 
     """
-    site = models.ForeignKey(Site)
-    date_for = models.DateField()
+    site = models.ForeignKey(Site, db_index=True)
```
OmarIthawi, line 209: As far as I know `ForeignKey` always implies db_index at the SQL level. [Stackoverflow agrees](https://stackoverflow.com/a/304339).

```suggestion
    site = models.ForeignKey(Site)
```

figures/models.py
```diff
@@ -206,11 +206,11 @@ class LearnerCourseGradeMetrics(TimeStampedModel):
...
+    date_for = models.DateField(db_index=True)
     # TODO: We should require the user
-    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
-    course_id = models.CharField(max_length=255, blank=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True, db_index=True)
```
OmarIthawi, line 212: Ditto.

```suggestion
    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True)
```

johnbaldwin:
@OmarIthawi Thanks for the comments. One item I forgot was running and adding the migration file. I'll inspect the SQL shell to see what indexes are already on the learner course grade metrics table and if the SO collective is right, then I'll remove those parameters

johnbaldwin:
No longer need this PR. Superseded by https://github.com/appsembler/figures/pull/288

PR #226, opened 2020-06-15
Updates pipeline enrollment metrics queries to improve performance

johnbaldwin:
This commit should dramatically improve the query performance for the
enrollment metrics pipeline

What was wrong?

Queries were very slow because of a 'LMIT 1' issues with MySQL. For a starting point, see here

    https://stackoverflow.com/questions/15460133/mysql-dramatically-slower-query-execution-if-use-limit-1-instead-of-limit-5

In Django, we were doing a filter query that returns a single record or
`None`. Examples:

```
StudentModule.objects.filter(**filter_args).latest('modified')
StudentModule.objects.filter(**filter_args).order_by('-modified).first()
```

Query functions such as `latest`, `first`, `last` and so on add a `LIMIT
1` to the underlying SQL query, which has apparent negative performance on the query analyzer

To address this, we do two things

1. For the specified course, we filter the StudentModule records
2. For the specifid learner in the course, we filter

Also, LearnerCourseGradesMetrics queries are slow as the model needs indexing
on fields including site, course, and learner. We address this twofold

1. We will add indexing to the needed fields after we prune old records.
This is so we're not indexing records we are just going to delete anyway
2. We filter all LearnerCourseGradeMetrics records for the specified
course

This commit performs #2 above to then filter from this queryset to find
LearnerCourseGradeMetrics records for the specified learner in the
course

Enrollment Metrics tests have been updated to reflect changes in the
production code

### DIFF
--- a/figures/pipeline/enrollment_metrics.py
+++ b/figures/pipeline/enrollment_metrics.py
@@ -56,8 +56,8 @@ from figures.models import LearnerCourseGradeMetrics, PipelineError
 from figures.pipeline.logger import log_error
 from figures.sites import (
     get_site_for_course,
+    get_student_modules_for_course_in_site,
     course_enrollments_for_course,
-    student_modules_for_course_enrollment,
     UnlinkedCourseError,
     )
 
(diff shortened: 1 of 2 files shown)


figures/pipeline/enrollment_metrics.py
```diff
@@ -144,17 +152,14 @@ def collect_metrics_for_enrollment(course_enrollment, date_for=None, **_kwargs):
...
 
     # The following are two different ways to avoide the dreaded error
     #     "Instance of 'list' has no 'order_by' member (no-member)"
     # See: https://github.com/PyCQA/pylint-django/issues/165
-    student_modules = student_modules_for_course_enrollment(ce=course_enrollment)
+    student_modules = course_sm.filter(
+        student_id=course_enrollment.user.id).order_by('-modified')
     if student_modules:
-        most_recent_sm = student_modules.latest('modified')
+        most_recent_sm = student_modules[0]
```
OmarIthawi, line 162: How many records are we filtering from?

johnbaldwin: @OmarIthawi The top five counts are 398, 382, 309, 309, 308.

As far as limiting the queries, AFAIK, there's no point in doing a "LIMIT 5" because querysets are lazy evaluated unless you use the "step" parameter of python slicing. Please see here: https://docs.djangoproject.com/en/1.11/topics/db/queries/#limiting-querysets

PR #229, opened 2020-06-22
Implement a new view: Learners Progress Overview

grozdanowski:
Per requests by our users, we found that we are missing a central point of overview for learner progress. That is why we are working on implementing this new view: Learners Progress Overview. It allows to view the progress of _all learners_ in _all courses_ at the _same time_. It also allows to filter out certain courses to only show the learners that have enrolled in them, as well as having a search enabled (by email, username or full name). After creating your filtered and sorted view, you can export the data from this view as a CSV on-the-fly from the frontend.

**Note: _while the view is ready for testing, we need to figure out how to test this out (especially CSV generation) in production against a serious amount of data. My recommendation is to temporarily remove the menu link for this new view and push it to production, then access it directly using the URL and test it out in real world. I'm also open for other suggestions._**

Screenshot 2020-06-22 at 18 08 55

### DIFF
--- a/figures/filters.py
+++ b/figures/filters.py
@@ -144,10 +144,15 @@ class UserFilterSet(django_filters.FilterSet):
         replaced with spaces, so we need to put the '+' back in for CourseKey
         to be able to create a course key object from the string
         '''
-        course_key = CourseKey.from_string(value.replace(' ', '+'))
-        enrollments = get_enrolled_in_exclude_admins(course_id=course_key)
-        user_ids = enrollments.values_list('user__id', flat=True)
-        return queryset.filter(id__in=user_ids)
+        total_user_ids = []
+        if value is not None:
+            courseIds = value.split('|')
+            for courseId in courseIds:
+                course_key = CourseKey.from_string(courseId.replace(' ', '+'))
+                enrollments = get_enrolled_in_exclude_admins(course_id=course_key)
+                user_ids = enrollments.values_list('user__id', flat=True)
+                total_user_ids = total_user_ids + list(set(user_ids) - set(total_user_ids))
+        return queryset.filter(id__in=total_user_ids)
 
 
 class CourseDailyMetricsFilter(django_filters.FilterSet):
--- a/frontend/src/views/ProgressOverview.js
+++ b/frontend/src/views/ProgressOverview.js
@@ -9,42 +9,40 @@ import HeaderContentStatic from 'base/components/header-views/header-content-sta
 import Paginator from 'base/components/layout/Paginator';
 import ListSearch from 'base/components/inputs/ListSearch';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faCheck, faAngleDoubleUp, faAngleDoubleDown } from '@fortawesome/free-solid-svg-icons';
+import { faAngleDoubleUp, faAngleDoubleDown } from '@fortawesome/free-solid-svg-icons';
+import { ExportToCsv } from 'export-to-csv';
+import { Multiselect } from 'multiselect-react-dropdown';
 
 import classNames from 'classnames/bind';
 let cx = classNames.bind(styles);
-
 
 class ProgressOverview extends Component {
   constructor(props) {
     super(props);
 
     this.state = {
-      coursesIndex: [],
-      coursesList: [],
       learnersList: [],
       selectedCourses: [],
+      coursesFilterOptions: [],
       perPage: 20,
       count: 0,
       pages: 0,
       currentPage: 1,
       searchQuery: '',
       ordering: 'profile__name',
+      selectedCourseIds: '',
+      csvExportProgress: 0,
     };
-
-    this.getUsers = this.getUsers.bind(this);
-    this.getCourses = this.getCourses.bind(this);
-    this.setCoursesIndex = this.setCoursesIndex.bind(this);
-    this.setPerPage = this.setPerPage.bind(this);
-    this.setSearchQuery = this.setSearchQuery.bind(this);
-    this.setOrdering = this.setOrdering.bind(this);
-    this.constructApiUrl = this.constructApiUrl.bind(this);
+    this.fetchFullViewData = this.fetchFullViewData.bind(this);
+    this.getUsersForCsv = this.getUsersForCsv.bind(this);
   }
 
-  constructApiUrl(rootUrl, searchQuery, orderingType, perPageLimit, resultsOffset) {
+  constructApiUrl = (rootUrl, searchQuery, selectedCourseIds, orderingType, perPageLimit, resultsOffset) => {
     let requestUrl = rootUrl;
     // add search term
     requestUrl += '?search=' + searchQuery;
+    // add course filtering
+    requestUrl += '&enrolled_in_course_id=' + selectedCourseIds;
     // add ordering
     requestUrl += '&ordering=' + orderingType;
     // add results per page limit
@@ -55,7 +53,7 @@ class ProgressOverview extends Component {
     return requestUrl;
   }
 
-  getCourses() {
+  getCourses = () => {
     const requestUrl = apiConfig.coursesIndex + '?limit=1000';
     trackPromise(
       fetch((requestUrl), { credentials: "same-origin" })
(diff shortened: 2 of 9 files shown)


johnbaldwin:
@grozdanowski So if I understand right, one suggestion you have is to hide this view with only a direct link? 

Would be handy if we could feature flag this for specific users for testing. 

Regardless, we should label this page as "Beta" after we've proved basic operational functionality on staging and we're into scalability testing on prod.

I'll comment specifically in the new filter code about the filter implementation

figures/filters.py
```diff
@@ -144,10 +144,15 @@ def filter_enrolled_in_course_id(self, queryset,
...
-        return queryset.filter(id__in=user_ids)
+        total_user_ids = []
+        if value is not None:
+            courseIds = value.split('|')
+            for courseId in courseIds:
+                course_key = CourseKey.from_string(courseId.replace(' ', '+'))
+                enrollments = get_enrolled_in_exclude_admins(course_id=course_key)
+                user_ids = enrollments.values_list('user__id', flat=True)
+                total_user_ids = total_user_ids + list(set(user_ids) - set(total_user_ids))
+        return queryset.filter(id__in=total_user_ids)
```
johnbaldwin, line 155: @grozdanowski  First, cheers for diving into Python and the back end of Figures! I really want to encourage you, 

but now for the comments:

# Style nit

I don't believe I've formally written contribution docs to identify style guidelines for Figures, pretty much we use Google's python style guide: https://google.github.io/styleguide/pyguide.html#316-naming

Which means please use `course_id` and `course_ids` instead of `courseId` and `courseIDs` for local var names

# Data safety

This introduces a potential latent vulnerability: Course ids from any site can be in the query string. so we can have users who belong to other sites in `total_user_ids`. Now this is _potential_ and not *realized* because `queryset` contains user ids only for this site (phew!), so we have a white list that will just ignore non-site user ids in `total_user_ids`. This work _for now_.  However it is latent because if this code is then refactored or reused without `queryset` behaving as it does or if some future query filtering drops `queryset`, or if someone likes this code and does copy/paste programming without proper understanding of site isolation, data leak vulnerability is introduced.

We can fix this by validating course ids belong to the site. One easy way to do this would be something like the following:

```
# Top of module
from figures.sites import get_course_keys_for_site
from django.contrib.sites.shortcuts get_current_site

# in our Filter method

param_course_keys = [as_course_key(val.replace(' ', '+')) for val in value.split('|')]
site = get_current_site(self.request)
if not set(param_course_keys) <= set(get_course_keys_for_site(site)):
    raise InvalidCourseIdError("some message")  # this would be a new exception class
else:
    # handle code normally
    ...
```

However, this brings us to point 'B':

# B Data correctness  / consistency and expectations and performance

Let's say User X is an admin in Course X and a learner in Course B. If this is called with Course A and Course B in the params, should we include or not include User X? Now, this is not a big deal, but we want to make sure we are clear on what we want with regard to including/excluding course staff.


Also, the code is making N SQL calls, one for each course. Now `get_enrolled_in_exclude_admins` is pretty fast, I tested it in prod with a big site and the range I got was 0.004 to 0.01 sec per call.

Not too bad, and even if we have a lot of courses in the query string, we can likely skip worrying about O(n) performance because in reality, the query string size is going to be a problem before we have any kind of worry about asymptotic analysis (https://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string)


So if we decide that it is ok to return course staff, we can then do some easy query optimization, like the following:


```
# Top of figures/filter.py module
from django.contrib.sites.shortcuts get_current_site

# in our Filter class method:

    course_keys = [as_course_key(val.replace(' ', '+')) for val in course_query_str.split('|')]

    if not set(course_keys) <= set(get_course_keys_for_site(site)):
        raise InvalidCourseError('Request gave us invalid course keys')

    ce = CourseEnrollment.objects.filter(course_id__in=course_keys)

    # `.distinct()` is probably not needed, might have additional overhead, but
    # adding it gets us a list without duplicate user_ids, since users can be
    # enrolled in multiple courses in a site
    return queryset.filter(id__in=ce.values('user_id').distinct())
```

If we DO need to exclude course staff unless there is a course for which the admin is enrolled as a learner, we can adapt the `get_enrolled_in_exclude_admins` to handle multiple courses so we've got one SQL call to retrieve user ids for the whole list of courses. We might need to do some RAW nested SQL to get there, but we'll do the cleanest code that is in the same order of magnitude performance.

johnbaldwin:
@grozdanowski I commented. Also I posted a PR in our `scripts` repo and added you as a reviewer. There I have demo code for doing user filtering. That can be a sandbox you can run with production data to test query performance. We can chat in slack on that or hop on a vid-chat

figures/filters.py
```diff
@@ -144,10 +144,15 @@ def filter_enrolled_in_course_id(self, queryset,
         replaced with spaces, so we need to put the '+' back in for CourseKey
         to be able to create a course key object from the string
         '''
-        course_key = CourseKey.from_string(value.replace(' ', '+'))
-        enrollments = get_enrolled_in_exclude_admins(course_id=course_key)
-        user_ids = enrollments.values_list('user__id', flat=True)
-        return queryset.filter(id__in=user_ids)
+        total_user_ids = []
+        if value is not None:
+            courseIds = value.split('|')
```
OmarIthawi, line 149:
```suggestion
            courseIds = value.split(',')
```

figures/filters.py
```diff
@@ -144,10 +144,15 @@ def filter_enrolled_in_course_id(self, queryset,
...
+            for courseId in courseIds:
+                course_key = CourseKey.from_string(courseId.replace(' ', '+'))
```
OmarIthawi, line 151:
```suggestion
                course_key = CourseKey.from_string(courseId)
```

frontend/src/views/ProgressOverview.js
```diff
@@ -94,43 +93,128 @@ class ProgressOverview extends Component {
...
     }, () => {
       this.getUsers();
     })
   }
 
+  onChangeSelectedCourses = (selectedList, item) => {
+    const selectedIdsList = selectedList.map((course, index) => {
+      return course.id;
+    });
+    const selectedCourseIds = selectedIdsList.join('|');
```
OmarIthawi, line 124:
```suggestion
    const selectedCourseIds = selectedIdsList.join(',');
```

frontend/src/views/ProgressOverview.js
```diff
@@ -9,42 +9,40 @@ import HeaderContentStatic from 'base/components/header-views/header-content-sta
...
+    this.getUsersForCsv = this.getUsersForCsv.bind(this);
   }
 
-  constructApiUrl(rootUrl, searchQuery, orderingType, perPageLimit, resultsOffset) {
+  constructApiUrl = (rootUrl, searchQuery, selectedCourseIds, orderingType, perPageLimit, resultsOffset) => {
     let requestUrl = rootUrl;
     // add search term
     requestUrl += '?search=' + searchQuery;
+    // add course filtering
+    requestUrl += '&enrolled_in_course_id=' + selectedCourseIds;
```
OmarIthawi, line 45:
```suggestion
    requestUrl += '&enrolled_in_course_id=' + encodeURIComponent(selectedCourseIds);
```

frontend/src/views/ProgressOverview.js
```diff
@@ -9,42 +9,40 @@ import HeaderContentStatic from 'base/components/header-views/header-content-sta
...
-    this.constructApiUrl = this.constructApiUrl.bind(this);
+    this.fetchFullViewData = this.fetchFullViewData.bind(this);
+    this.getUsersForCsv = this.getUsersForCsv.bind(this);
   }
 
-  constructApiUrl(rootUrl, searchQuery, orderingType, perPageLimit, resultsOffset) {
+  constructApiUrl = (rootUrl, searchQuery, selectedCourseIds, orderingType, perPageLimit, resultsOffset) => {
     let requestUrl = rootUrl;
     // add search term
     requestUrl += '?search=' + searchQuery;
```
OmarIthawi, line 43:
```suggestion
    requestUrl += '?search=' + ecodeURIComponent(searchQuery);
```

frontend/src/views/ProgressOverview.js
```diff
@@ -9,42 +9,40 @@ import HeaderContentStatic from 'base/components/header-views/header-content-sta
...
+    // add course filtering
+    requestUrl += '&enrolled_in_course_id=' + selectedCourseIds;
     // add ordering
     requestUrl += '&ordering=' + orderingType;
```
OmarIthawi, line 47:
```suggestion
    requestUrl += '&ordering=' + encodeURIComponent(orderingType);
```

grozdanowski:
This PR is to me dismissed, as per new work replacing this one. New PR is here: [https://github.com/appsembler/figures/pull/240](https://github.com/appsembler/figures/pull/240)

PR #233, opened 2020-07-01
New endpoint for enrollment metrics - provide data captured in LCGM

johnbaldwin:
Added new endpoint to retrieve enrollment metrics. 

The new base endpoint is `/figures/api/enrollment-metrics/`which provides an API view for `LearnerCourseGradeMetrics` model. The `LearnerCourseGradeMetrics` model is where we capture learner progress and completion data. These data originate from `CourseGrade` and populated during the daily pipeline jobs with calls to `CourseGradeFactory().read(...)`

* Updated LearnerCourseGradeMetricsManager model manager to retrieve "course completed" data. 
* Added broad query parameter filtering capability to this endpoint. All fields should be filterable except for `site`. Also added two list filters `user_ids=` and `course_ids=` which take comma delimited lists of user ids and course ids respectively. Added two filters `only_completed` and `exclude_completed` that take a `True`. _Yes, Capitalize the first letter, this is an artifact of django-filter. We can improve this in the future, for now, just want to have the functionality in so we can start working with it in the front end_

* Added experimental raw SQL queries to the LCGM model manager. These are not called by the endpoint and here to test retrieving the most recent LCGM record for a set of users

# Fixes
- removed "LIMIT 1" query from  LCGM model manager `most_recent_for_learner_course` method
- Added pylint: disable=E1101 to stop pylint failures in recognizing that `filter` is a method of the queryset (PyLint thinks the object is a list at these places).

We can retrieve the core "completed" metrics data now, but we're likely going to want to rework and/or improve the endpoint after we cleanup LCGM and add indexes to the `course_id` and `date_for` fields. We may also want to convert LCGM `completed` proper to a table column so we can index on that

Tests are mostly there, but need more coverage. First I wanted to get this out to review


johnbaldwin:
I'm working on improving coverage. Looks like the big spots are the new view, filter methods, and a couple of spots in the model update

johnbaldwin:
Coverage dropped into unhappy territory after merging from master. Working on boosting it above the waterline

johnbaldwin:
Added test coverage for cases where paginated results not retrieved for the "completed" and "completed_ids" actions

PR #234, opened 2020-07-12
Improve Performance for monthly active users API endpoint

johnbaldwin:
This PR is to improve the performance of Figures API to improve Figures UI page loading and prevent timeouts on retrieving monthly active users, which gets slow as we're currently pulling live data for the current month

Figures currently performs a live query on the `edx-platform`
`courseware.models.StudentModule` model, but the query is too slow when there are many `StudentModule` records. The rough order of magnitude (ROM) for "many" appears to be about 10^6.

There are two commits.

* First commit updates the data model
* Second commit updates the pipeline and metrics code

The data model update adds 'mau' field to 'SiteDailyMetrics' model

The pipeline code now populates 'SiteDailyMetrics.mau' field with the latest numbers for the month as of when the pipeline is run.

The slow query causes an API timeout, degrading performance for Figures UI.

This commit provides the storage for this metric so the following commit can update the pipeline code and the 3rd commit to update the API code

Removed the old MAU calculation code which is now dead code and removed the tests


figures/metrics.py
```diff
@@ -273,12 +275,17 @@ def get_site_mau_current_month(site):
...
+    to do this query in the pipeline, storing values to `SiteDailyMetrics.mau`
+
+    TODO: Move this method to figures.mau
     """
     month_for = datetime.datetime.utcnow()
     site_sm = figures.sites.get_student_modules_for_site(site)
     curr_sm = site_sm.filter(modified__year=month_for.year,
                              modified__month=month_for.month)
-    return curr_sm.values('student__id').distinct().count()
+    return curr_sm.values('student__id').distinct()
```
johnbaldwin, line 288: We do this so we're able to use this function to retrieve the distinct ids, then we can get the count from the querset for when we need the count

OmarIthawi:
The commit message says that there should be a third commit @johnbaldwin? Is that still correct? When can I review this PR?

<kbd>image</kbd>

johnbaldwin:
https://github.com/appsembler/figures/pull/234#issuecomment-657367587

@OmarIthawi I didn't need to do the 3rd commit, which was going to update the viewset. Switching logic at `figures.metrics` took care of it, thanks to abstraction

johnbaldwin:
Thanks Omar!

PR #239, opened 2020-08-01
Learner metrics endpoint - Prerelease for evaluation

johnbaldwin:
This PR is for the initial evaluation version of the learner metrics endpoint for Matej's implementation on the front end and evaluation in production.

The endpoint is `/figures/api/learner-metrics/`

* There is a basic viewset just to exercise the code. The test requires test data to be filled out and tested in the response

* UserFilterSet needs to be updated or an alternate filter set needs to be used in order to provide more filtering, in particular
 * Show only users who have enrollments
 * Show only users who do not have enrollments
 * Show only users who have completed
 * Show only users who have not completed

* List serializers need to be added to prefetch data to improve API performance
* test_learner_metrics_viewset needs to be completed
* Updated the CourseEnrollment mock to provide the `is_enrolled` method

### DIFF
--- a/tests/views/test_learner_metrics_viewset.py
+++ b/tests/views/test_learner_metrics_viewset.py
@@ -1,0 +1,165 @@
+"""Tests Figures learner-metrics viewset
+"""
+
+import pytest
+
+import django.contrib.sites.shortcuts
+from rest_framework import status
+from rest_framework.test import APIRequestFactory, force_authenticate
+
+from figures.sites import get_user_ids_for_site
+from figures.views import LearnerMetricsViewSet
+
+from tests.factories import (
+    CourseEnrollmentFactory,
+    CourseOverviewFactory,
+    # LearnerCourseGradeMetricsFactory,
+    OrganizationFactory,
+    SiteFactory,
+    UserFactory,
+)
+
+from tests.helpers import organizations_support_sites
+from tests.views.base import BaseViewTest
+from tests.views.helpers import is_response_paginated
+
+if organizations_support_sites():
+    from tests.factories import UserOrganizationMappingFactory
+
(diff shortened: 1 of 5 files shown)


johnbaldwin:
I need to fix the unit test. It is failing because it is getting all users instead of the expected users. Should still work in devsite. I'll work on that tomorrow

tests/views/test_learner_metrics_viewset.py
```diff
@@ -0,0 +1,156 @@
...
+                                     site=site,
+                                     caller=caller,
+                                     action='list')
+
+        assert response.status_code == status.HTTP_200_OK
+        assert is_response_paginated(response.data)
+        results = response.data['results']
+        # Check user ids
+        result_ids = [obj['id'] for obj in results]
+        assert set(result_ids) == set([obj.user_id for obj in enrollments]+[caller.id])
```
johnbaldwin, line 152: This is what is failing because the result_ids is also including users created in the base test class from which this class inherits

tests/views/test_learner_metrics_viewset.py
```diff
@@ -0,0 +1,156 @@
...
+                    "progress_details": {
+                        "sections_worked": 20,
+                        "points_possible": 100.0,
+                        "sections_possible": 20,
+                        "points_earned": 50.0
+                    }
+                }
+            ]
+        }
+    ```
```
johnbaldwin, line 94: @grozdanowski I pasted a sample from the mock results in devsite

tests/views/test_learner_metrics_viewset.py
johnbaldwin: This is fixed

PR #248, opened 2020-08-24
John/learner metrics multi course filtering

johnbaldwin:
This PR adds multiple course filtering for the `learner-metrics` endpoint.

# New functions to support filtering 

Adds two functions to figures.sites. This is to support multi-course
filtering for the learner-metrics endpoint

1. enrollments_for_course_ids

This retrieves a CourseEnrollment queryset for enrollments that belong
to the list of course ids argument

2. users_enrolled_in_courses

This retrieves a User queryset for users that are enrolled in the
courses identified by the course ids argument

Added tests for the new functionality

# New learner-metrics query parameter, "course=<course-id>" 

The `learner-metrics` endpoint now filters on one or more courses using
'course=some-course-id` query parameters. To filter on multiple courses,
simply do multiple parameters of the same "course" key. Example

```
?course-my-first-course-id&course=my-second-course-id
```

* Added query param tests to the learner metrics viewset test class

### DIFF
--- a/figures/views.py
+++ b/figures/views.py
@@ -420,14 +420,29 @@ class LearnerMetricsViewSet(CommonAuthMixin, viewsets.ReadOnlyModelViewSet):
     search_fields = ['username', 'email', 'profile__name']
     ordering_fields = ['username', 'email', 'profile__name', 'is_active', 'date_joined']
 
+    def query_param_course_keys(self):
+        """
+        TODO: Mixin this
+        """
+        cid_list = self.request.GET.getlist('course')
+        return [CourseKey.from_string(elem.replace(' ', '+')) for elem in cid_list]
+
     def get_queryset(self):
+        """
+        This function has a hack to filter users until we can get the `filter_class`
+        working
+        """
         site = django.contrib.sites.shortcuts.get_current_site(self.request)
         queryset = figures.sites.get_users_for_site(site)
+        course_keys = self.query_param_course_keys()
+        if course_keys:
+            queryset = figures.sites.users_enrolled_in_courses(course_keys)
         return queryset
 
     def get_serializer_context(self):
         context = super(LearnerMetricsViewSet, self).get_serializer_context()
         context['site'] = django.contrib.sites.shortcuts.get_current_site(self.request)
+        context['course_keys'] = self.query_param_course_keys()
         return context
 
 
--- a/figures/serializers.py
+++ b/figures/serializers.py
@@ -888,4 +888,8 @@ class LearnerMetricsSerializer(serializers.ModelSerializer):
         site_enrollments = figures.sites.get_course_enrollments_for_site(
             self.context.get('site'))
         user_enrollments = site_enrollments.filter(user=user)
+        course_keys = self.context.get('course_keys')
+        if course_keys:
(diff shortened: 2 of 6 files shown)


grozdanowski (approved):
Thanks, John! I have tested this out locally and it works like a charm. I also made changes required to the frontend (handling of multi-course filtering in API URL generation) and tested them out. This looks ready for staging to me!

figures/views.py
```diff
@@ -420,14 +420,29 @@ class LearnerMetricsViewSet(CommonAuthMixin, viewsets.ReadOnlyModelViewSet):
     search_fields = ['username', 'email', 'profile__name']
     ordering_fields = ['username', 'email', 'profile__name', 'is_active', 'date_joined']
 
+    def query_param_course_keys(self):
+        """
+        TODO: Mixin this
+        """
+        cid_list = self.request.GET.getlist('course')
+        return [CourseKey.from_string(elem.replace(' ', '+')) for elem in cid_list]
```
OmarIthawi, line 428: The `.replace(' ', '+')` is an anti pattern in the backend and is error-prune. The frontend should send properly escaped course keys with `encodeURIComponent` or use `fetch()` to take care of that.

```suggestion
    def query_param_course_keys(self):
        """
        TODO: Mixin this
        """
        cid_list = self.request.GET.getlist('course')
        return [CourseKey.from_string(elem) for elem in cid_list]
```

OmarIthawi (approved):

figures/views.py
johnbaldwin: I"ll make this change right after I have some coffee :)

johnbaldwin: Well, I would love to make this change, but when I try it, I get this. So the anti-pattern appears to be required still.

```
>>> cid = 'course-v1:StarFleetAcademy SFA01 2161'
>>> from opaque_keys.edx.keys import CourseKey
>>> CourseKey.from_string(cid)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/jbaldwin/.virtualenvs/figtahoe2/lib/python2.7/site-packages/opaque_keys/__init__.py", line 197, in from_string
    return cls.deprecated_fallback._from_deprecated_string(serialized)
  File "/Users/jbaldwin/.virtualenvs/figtahoe2/lib/python2.7/site-packages/opaque_keys/edx/locator.py", line 379, in _from_deprecated_string
    raise InvalidKeyError(cls, serialized)
InvalidKeyError: <class 'opaque_keys.edx.locator.CourseLocator'>: course-v1:StarFleetAcademy SFA01 2161
```

OmarIthawi: @johnbaldwin that's right, `course-v1:StarFleetAcademy SFA01 2161` is _not_ a valid key. Therefore the frontend should not send unescaped which what would make it an invalid key.

It's as if the frontend keeps adding a 999 to numbers and asking the API to subtract 999 before using the `page` argument. This will be more of an issue in the cases like:

 - One of our customers tries to use the figures API through a library that escapes the parameters properly like Python `requests` or pretty much every HTTP client out there.
 - Or if it's us working with Figures APIs through such libraries.

It's bothering that we're doing an extra step every time we make a Figures API to fix something in the _wrong_ place.

PR #255, opened 2020-09-10
Fixes to Learner Progress Overview

grozdanowski:
Here's a list of items fixed:

**If you select a course, you can’t add another course with the same name to the dashboard** 
Now you can and you can easily distinguish them because:
- in selector now you have additional information to easily identify the run
- in data render list there’s now also course id to easily identify the run

**Data is shown to a needless level of specificity, causing fields to overflow**
Fixed. Literally

**Data that is 0 is invisible, just leaving slashes**
This is now fixed - checks added into code.

**The page looks weird when you only have one course selected**
Not anymore! Now introducing - “wide view toggle”. Initially you have regular screen width. If you need more, you can toggle it.

**Loading all courses initially in LPV is useless**
New behaviour:
- initially no data is loaded to render
- data will be rendered only if there is a search query and/or courses selected

**Add more info to multicourse select dropdown to distinguish reruns**
Done.

**Enable course multiselector search by course number**
Done.

### DIFF
--- a/frontend/src/views/ProgressOverview.js
+++ b/frontend/src/views/ProgressOverview.js
@@ -32,6 +32,7 @@ class ProgressOverview extends Component {
       ordering: 'profile__name',
       selectedCourseIds: '',
       csvExportProgress: 0,
+      wideView: false,
     };
     this.fetchFullViewData = this.fetchFullViewData.bind(this);
     this.getUsersForCsv = this.getUsersForCsv.bind(this);
@@ -100,7 +101,7 @@ class ProgressOverview extends Component {
     this.setState({
       perPage: newValue,
     }, () => {
-      this.getUsers();
+      (this.state.selectedCourses.length || this.state.searchQuery) && this.getUsers();
     })
   }
 
@@ -108,7 +109,7 @@ class ProgressOverview extends Component {
     this.setState({
       searchQuery: newValue
     }, () => {
-      this.getUsers();
+      (this.state.selectedCourses.length || this.state.searchQuery) && this.getUsers();
     })
   }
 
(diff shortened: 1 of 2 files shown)


frontend/src/views/ProgressOverview.js
```diff
@@ -196,7 +196,14 @@ class ProgressOverview extends Component {
       coursesFilter.forEach((course, i) => {
         const userProgress = userCoursesImmutable.find(singleCourse => singleCourse.get('course_id') === course.id);
         if (userProgress) {
-          singleRecord[course.id] = `Progress: ${userProgress.getIn(['progress_percent'])}/1 | Sections: ${userProgress.getIn(['progress_details', 'sections_worked'])}/${userProgress.getIn(['progress_details', 'sections_possible'])} | Points: ${userProgress.getIn(['progress_details', 'points_earned'])}/${userProgress.getIn(['progress_details', 'points_possible'])}`;
+          
+          const progressPercent = (userProgress.getIn(['progress_percent'])) ? userProgress.getIn(['progress_percent']).toFixed(2) : '-';
```
OmarIthawi, line 202: This could be a helper function, but okay for now.

OmarIthawi (approved):
Fun fact: LGTM.com is an company site that advocates against "LGTM" with a [21 million dollar investment](https://blog.semmle.com/semmle-launches-globally-series-b/).

But hey, LGTM 😁

frontend/src/views/ProgressOverview.js
grozdanowski: Yup, agreed. However this whole view is done for now in a kind of anti-pattern approach. It is a monolith, and should be refactored down the line.

PR #257, opened 2020-09-10
Create codeql-analysis.yml

thraxil:
Enable automatic security scanning on the repo.

Pretty neat that on this one it picked up that there was JS and runs a scan on that as well.


johnbaldwin (approved):
Cool, thanks Anders!

PR #261, opened 2020-09-26
Fix code that doesn't work on Ginkgo (Django 1.8)

johnbaldwin:
https://appsembler.atlassian.net/browse/RED-1348

The issue is that Django 1.8 does not support relative comparison for
the 'day' field in QuerySets. Example, this does not work:

`SomeModel.objects(my_date_field__day__lte=some_day`

So we need to perform an explicit datetime range when running with
Django 1.8

Since doing the commit as specified for "non-ginkgo" is more readable,
we don't want to throw it away. Instead, when we no longer have to
support Django 1.8 (meaning Ginkgo), we can then remove this check in
MAU

Tests updated to improve date range checking for MAU

Also added missing `django-celery` to the ginkgo requirements file

### DIFF
--- a/figures/mau.py
+++ b/figures/mau.py
@@ -2,7 +2,10 @@
 This module provides MAU metrics retrieval functionality
 """
 
-from datetime import datetime
+from datetime import datetime, timedelta
+from django.utils.timezone import utc
+
+from figures.compat import RELEASE_LINE
 
 from figures.models import CourseMauMetrics, SiteMauMetrics
 from figures.sites import (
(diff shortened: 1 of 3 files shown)


figures/mau.py
```diff
@@ -84,11 +87,39 @@ def mau_1g_for_month_as_of_day(sm_queryset, date_for):
...
     who are active in the site
 
+    Retrieves records based on date of the `StudentModule.modified` field
     Returns a queryset of distinct user ids
     """
-    month_sm = sm_queryset.filter(modified__year=date_for.year,
-                                  modified__month=date_for.month,
-                                  modified__day__lte=date_for.day)
+
+    # For now, directly doing compat check here and not creating an abstracted
```
OmarIthawi, line 94:
```suggestion
    # TODO: Remove this `if` branch after dropping Ginkgo support.
    # For now, directly doing compat check here and not creating an abstracted
```

OmarIthawi (approved):

figures/mau.py
johnbaldwin: Sounds good. Thanks

PR #277, opened 2020-10-28
Fix the comparison showing N/A when comparing to zero

grozdanowski:
As per title. Essentially it would check if the past value exists, however it treated 0 (zero) the same way as undefined or Null. Now fixed.


johnbaldwin:
@grozdanowski You can't add the compiled assets to master. Do you need me to fix your commits?

johnbaldwin (changes requested):
We need to remove compiled assets

johnbaldwin:
I killed travis so it wouldn't be in the queue, there's no python in this PR, and the travis delays are slowing us down

grozdanowski:
@johnbaldwin compiled assets removed.

OmarIthawi (approved):

grozdanowski:
@johnbaldwin changes made, just waiting on your approval and you can merge it.

johnbaldwin (approved):
Thanks for removing the assets, @grozdanowski !

johnbaldwin:
@grozdanowski I merged master in here and will merge this PR into master after the tests finish

PR #289, opened 2020-12-03
Use the Open edX named releases to conditionally import platform objects

johnbaldwin:
This commit exists to address the problem that a RuntimeError can be
raised instead of an ImportError when importing objects from
ed-platform.

The options to fix are check for ImportError AND RuntimeError. However,
we opt (at least for now) to check the Open edX named release to
conditionally import objects when the namespace has changed between
named releases.

Here is an example of the error when only ImportError is handled:
```
RuntimeError: Conflicting 'studentmodule' models in application 'courseware': <class 'courseware.models.StudentModule'> and <class 'lms.djangoapps.courseware.models.StudentModule'>.
```

NOTE: This may break Juniper compat as it was not tested in a live
Juniper devstack

NOTE: See @OmarIthawi 's comments and this issue: https://github.com/appsembler/figures/issues/290

### DIFF
--- a/figures/compat.py
+++ b/figures/compat.py
@@ -11,46 +11,45 @@ a lowercase string, such as 'ginkgo' or 'hawthorn'
 # pylint: disable=ungrouped-imports,useless-suppression
 
 from __future__ import absolute_import
+
+
+class UnsuportedOpenedXRelease(Exception):
+    pass
+
+
+# Pre-Ginkgo does not define `RELEASE_LINE`
 try:
     from openedx.core.release import RELEASE_LINE
 except ImportError:
-    # we are pre-ginkgo
-    RELEASE_LINE = None
-
-try:
-    # First try to import for the path as of hawthorn
-    from lms.djangoapps.grades.course_grade_factory import CourseGradeFactory
-except ImportError:
-    try:
-        # Next try to import for the path as of ginkgo
-        from lms.djangoapps.grades.new.course_grade_factory import CourseGradeFactory
-    except ImportError:
-        # try the pre-ginkgo path
-        from lms.djangoapps.grades.new.course_grade import CourseGradeFactory    # noqa: F401
+    raise UnsuportedOpenedXRelease(
+        'Unidentified Open edX release: '
+        'figures.compat could not import openedx.core.release.RELEASE_LINE')
 
 
 if RELEASE_LINE == 'ginkgo':
+    from lms.djangoapps.grades.new.course_grade_factory import CourseGradeFactory  # noqa pylint: disable=unused-import,import-error
+else:  # Assume Hawthorn or greater
+    from lms.djangoapps.grades.course_grade_factory import CourseGradeFactory  # noqa pylint: disable=unused-import,import-error
+
+if RELEASE_LINE == 'ginkgo':
     from certificates.models import GeneratedCertificate  # noqa pylint: disable=unused-import,import-error
-else:
+else:  # Assume Hawthorn or greater
     from lms.djangoapps.certificates.models import GeneratedCertificate  # noqa pylint: disable=unused-import,import-error
 
-
-try:
-    from lms.djangoapps.courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
-except ImportError:
-    # Backward compatibily for pre-Juniper releases
+if RELEASE_LINE in ['ginkgo', 'hawthorn']:
     from courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
+else:  # Assume Juniper or greater
+    from lms.djangoapps.courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
 
-try:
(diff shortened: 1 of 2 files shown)


figures/compat.py
```diff
@@ -11,46 +11,35 @@
...
+if RELEASE_LINE == 'juniper':
     from lms.djangoapps.courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
-except ImportError:
-    # Backward compatibily for pre-Juniper releases
+else:  # Backward compatibily for pre-Juniper releases
     from courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
 
-try:
+if RELEASE_LINE == 'juniper':
     from lms.djangoapps.courseware.courses import get_course_by_id  # noqa pylint: disable=unused-import,import-error
```
OmarIthawi, line 35: Should work fine until Koa.

OmarIthawi (approved):

figures/compat.py
```diff
@@ -11,46 +11,35 @@
...
-try:
+if RELEASE_LINE == 'juniper':
     from lms.djangoapps.courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
-except ImportError:
-    # Backward compatibily for pre-Juniper releases
+else:  # Backward compatibily for pre-Juniper releases
     from courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
 
-try:
+if RELEASE_LINE == 'juniper':
```
OmarIthawi, line 34: Should work fine in Koa.

```suggestion
if RELEASE_LINE in ['juniper', 'koa']:
```

figures/compat.py
```diff
@@ -11,46 +11,35 @@
...
 
 if RELEASE_LINE == 'ginkgo':
     from certificates.models import GeneratedCertificate  # noqa pylint: disable=unused-import,import-error
-else:
+else:  # Assume Hawthorn or greater
     from lms.djangoapps.certificates.models import GeneratedCertificate  # noqa pylint: disable=unused-import,import-error
 
-
-try:
+if RELEASE_LINE == 'juniper':
```
OmarIthawi, line 29:
```suggestion
if RELEASE_LINE in ['juniper', 'koa']:
```

```
# koa
lms/djangoapps/instructor_analytics/basic.py:from lms.djangoapps.courseware.models import StudentModule
```

figures/compat.py
```diff
@@ -11,46 +11,35 @@
...
     from lms.djangoapps.courseware.courses import get_course_by_id  # noqa pylint: disable=unused-import,import-error
-except ImportError:
-    # Backward compatibily for pre-Juniper releases
+else:  # Backward compatibily for pre-Juniper releases
     from courseware.courses import get_course_by_id  # noqa pylint: disable=unused-import,import-error
 
-try:
-    from opaque_keys.edx.django.models import CourseKeyField  # noqa pylint: disable=unused-import,import-error
-except ImportError:
+if RELEASE_LINE == 'ginkgo':
```
OmarIthawi, line 39: This is future-proof. Great!

figures/compat.py
```diff
@@ -11,46 +11,35 @@
...
-    # Backward compatibily for pre-Juniper releases
+else:  # Backward compatibily for pre-Juniper releases
     from courseware.models import StudentModule  # noqa pylint: disable=unused-import,import-error
 
-try:
+if RELEASE_LINE == 'juniper':
     from lms.djangoapps.courseware.courses import get_course_by_id  # noqa pylint: disable=unused-import,import-error
-except ImportError:
-    # Backward compatibily for pre-Juniper releases
+else:  # Backward compatibily for pre-Juniper releases
```
OmarIthawi, line 36: This import is not future proof. 

Those imports would work but are not future-proof. To fix it needs to be restructured in the following form from the oldest to the most recent release:

```python
if ginkgo or hawthorn:
  import # old import
else:   # all future releases
  import  # future proof
```

OmarIthawi (approved):
I forgot to specify that all my previous comments where non-blockers. LGTM. Please feel free to merge. Also, thanks for making the issue #290.

johnbaldwin:
@OmarIthawi Thanks and I'm merging this in. Take another look. I believe I addressed your comments. Only thing I think needs to be addressed is getting test coverage for `figures.compat.course_grades`. See my git commit message for details

figures/compat.py
```diff
@@ -11,46 +11,45 @@
...
+    pass
+
+
+# Pre-Ginkgo does not define `RELEASE_LINE`
 try:
     from openedx.core.release import RELEASE_LINE
 except ImportError:
-    # we are pre-ginkgo
-    RELEASE_LINE = None
+    raise UnsuportedOpenedXRelease(
```
OmarIthawi, line 24: That's good!

OmarIthawi:
Thanks @johnbaldwin! I think this is pretty good and probably don't need the additional coverage.

PR #292, opened 2020-12-06
Add API docs generation to Figures devsite

johnbaldwin:
This was my Sunday afternoon project. I wanted to see what Swagger generated docs would look like for Figures.

OpenAPI/Swagger documentation generation now included in Figures
devsite. This requires running Figures devsite in "Juniper mode" with
Python 3.8 (3.5 might work but not tested).

See the file: `docs/soruce/api.rst` for details

Also added some improvements to the environment settings

* Moved the `.env` location to the same directory as `devsite/manage.py`
isntead of being in the devsite main package.
* Updated `.env.example`
* Added more settings to be enabled by `.env`

See the comments in `.env.example` for details


devsite/devsite/settings.py
```diff
@@ -16,16 +16,20 @@
...
 
+DEVSITE_BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+PROJECT_ROOT_DIR = os.path.dirname(DEVSITE_BASE_DIR)
 
 env = environ.Env(
+    DEBUG=(bool, True),
+    OPENEDX_RELEASE=(str, 'HAWTHORN'),
     FIGURES_IS_MULTISITE=(bool, False),
+    ENABLE_DEVSITE_CELERY=(bool, True),
+    ENABLE_OPENAPI_DOCS=(bool, False),
```
OmarIthawi, line 27: I'd lean to enable it by default in `devstie/settings.py` but not in prod.

```suggestion
    ENABLE_OPENAPI_DOCS=(bool, True),
```

OmarIthawi (approved):

devsite/devsite/settings.py
johnbaldwin: @OmarIthawi Sure, I'll do that after I refactor juniper requirements again so that we have only the needed packages installed for the juniper tests. And there is no "prod" for Figures devsite. Right now it is a pure dev site

johnbaldwin: I created an issue to make a mention: https://github.com/appsembler/figures/issues/294

PR #299, opened 2020-12-08
Increase waiting time upon typing in LPO search

grozdanowski:
Increase from 1 second to 2 seconds.


johnbaldwin (approved):
Cool, thanks @grozdanowski !

PR #301, opened 2020-12-09
Figures 0.4 learner progress overview performance boost

johnbaldwin:
* Sorry this is big, I'm in a rush and need to make sure the parts work, but I've layered the PR into individual commits focused on functional areas
* The PR commit messages should be pretty detailed

The core idea is that we have a new model, `figures.models.EnrollmentData` that can be retrieved with a `prefetch_related` call so there is only one additional query than querying the users (which is the first query). I still have query tweaking to do, but want to see
A. How this performs on staging
B. Check the data are correct (by comparing to the current endpoint which I moved to `/figures/api/learner-metrics-v1/`

I'm testing this PR's branch on staging (well, about to).

This is mostly complete. What's left

1. Backfilling EnrollmentData for existing CourseEnrollment and LearnerCourseGradeMetrics (LCGM) records

I'm probably going to do a Django management command to do a data migration. This way it gets triggered automatically on installation. Right now manual work.

2. The EnrollmentData records need to get updated when there is a new CourseEnrollment or a new LCGM record. I'm probably going to implement this with Django Signals to trigger the EnrollmentData record update 

https://appsembler.atlassian.net/browse/RED-1689


figures/migrations/0015_add_enrollment_data_model.py
```diff
@@ -0,0 +1,53 @@
...
+    if DJANGO_VERSION[0:2] == (1,8):
+        dependencies = [
+            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+            ('sites', '0001_initial'),
+            ('figures', '0014_add_indexes_to_daily_metrics'),
+        ]
+    else:  # Assuming 1.11+
+        dependencies = [
+            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+            ('sites', '0002_alter_domain_unique'),
```
OmarIthawi, line 24: This should work, no need to specify the latest migration from `sites`, so if you'd like it's possible to remove this `if` statement

```suggestion
            ('sites', '0001_initial'),
```

figures/management/commands/update_figures_enrollment_data.py
```diff
@@ -0,0 +1,59 @@
...
+    def handle(self, *args, **options):
+        print('BEGIN: Update Figures EnrollmentData')
+
+        if options['site']:
+            sites = [get_site(options['site'])]
+        else:
+            # Would be great to be able to filter out dead sites
+            # Would be really great to be able to filter out dead sites
+            # Would be really Really great to be able to filter out dead sites
+            # Would be really Really REALLY great to be able to filter out dead sites
```
OmarIthawi, line 49: I get you John. One idea is to soft delete: https://appsembler.atlassian.net/wiki/spaces/RT/blog/2020/12/10/1160708308/Soft+deleting+dead+sites

cc: @ahmedaljazzar

OmarIthawi (approved):
Thanks @johnbaldwin. It's pretty big for me to review it. I think we can merge once tests passes and depend on staging tests this time instead of an extensive code review.

OmarIthawi (approved):
Thanks again @johnbaldwin. My other comment had some random characters, I don't know why.

It's pretty big for me to review it, which is very understandable given the performance issues we're having. I think we can merge once tests passes and depend on staging tests this time instead of an extensive code review.

PR #310, opened 2020-12-15
Fix handling when general course data has null metrics

johnbaldwin:
When a course is created after the last daily pipline then there is not
a CourseDailyMetrics record for the course. This then causes the
'courses/general' API endpoint to have `null` for the 'metrics' key.

This commit is one way to address that. What it does is in the client's
courses page check if there is a value for metrics and set the values to
'N/A' if the value is not there.

There are other solutions, such as provide default '0' values in the
metrics code. But this way, we know that there are not metrics values
instead of metrics collected with zero value, which gives some clue that
there is an issue that needs to be investigated

### DIFF
--- a/figures/serializers.py
+++ b/figures/serializers.py
@@ -300,6 +300,15 @@ class GeneralCourseDataSerializer(serializers.Serializer):
             return []
 
     def get_metrics(self, obj):
+        """
+        Will return None if there is not a CourseDailyMetrics model
+
+        This will happen for courses that are created after the last daily
+        pipeline ran
+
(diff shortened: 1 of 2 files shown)


johnbaldwin:
@grozdanowski In case you don't get to reviewing this before I push it out, if you're not happy with what I did to fix the error on the front end, feel free to fix it how you wish. I would like to give a clue in the UI if there are missing metrics data rather than zero'd out "initial state" metrics data (initial state meaning when a course is first created), but your call

melvinsoft (approved):
@johnbaldwin Ship it, if @grozdanowski has a better solution, we can change it later.

PR #312, opened 2020-12-17
Fix bug and Improve error handling in Figures tasks and pipeline

johnbaldwin:
1. Fixed the bug, poorly placed parentheses within exception handling in the new code

2. Add pipeline exception handling - course daily metrics

The pipeline function, bulk_calculate_course_progress_data, does a lot
of work and a potential point of failure to fail all the metrics for the
course for the day on collecting just one metric.

So this commit traps for exceptions when the function
"bulk_calculate_course_progress_data" is called and sets the course daily
metrics 'average_progress' to None if it fails. Being 'None' helps
identify pipeline erors rather than setting progress to zero.

Added test to make sure we are logging when it fails

3. Add figures.tasks exception handling cdm and enrollment data
Added exception handling and logging to populate_single_cdm, the call to
update_enrollment_data in 'populate_daily_metrics' and exception
handling for each site so that one site does not fail the remaining
sites in the chain

* Updated tests to handle when update_enrollment_data raises an exception
* Updated tests to handle when a site level exception is raised

https://appsembler.atlassian.net/browse/RED-1709

### DIFF
--- a/figures/backfill.py
+++ b/figures/backfill.py
@@ -67,7 +67,7 @@ def backfill_enrollment_data_for_site(site):
             enrollment_data.append((obj, created))
         except CourseNotFound:
             errors.append('CourseNotFound for course "{}". '
-                          ' CourseEnrollment ID='.format(str(rec.course_id,
-                                                         rec.id)))
+                          ' CourseEnrollment ID='.format(str(rec.course_id),
+                                                         rec.id))
 
     return dict(results=enrollment_data, errors=errors)
--- a/figures/backfill.py
+++ b/figures/backfill.py
@@ -66,8 +66,9 @@ def backfill_enrollment_data_for_site(site):
                 course_id=rec.course_id)
             enrollment_data.append((obj, created))
         except CourseNotFound:
-            errors.append('CourseNotFound for course "{}". '
-                          ' CourseEnrollment ID='.format(str(rec.course_id),
-                                                         rec.id))
+            msg = ('CourseNotFound for course "{course}". '
+                   ' CourseEnrollment ID={ce_id}')
+            errors.append(msg.format(course=str(rec.course_id),
+                                     ce_id=rec.id))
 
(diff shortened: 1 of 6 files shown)


thraxil (approved):

johnbaldwin:
Thanks @thraxil !

figures/backfill.py
```diff
@@ -67,7 +67,7 @@ def backfill_enrollment_data_for_site(site):
             enrollment_data.append((obj, created))
         except CourseNotFound:
             errors.append('CourseNotFound for course "{}". '
-                          ' CourseEnrollment ID='.format(str(rec.course_id,
-                                                         rec.id)))
+                          ' CourseEnrollment ID='.format(str(rec.course_id),
+                                                         rec.id))
```
OmarIthawi, line 71: An additional `{}` is needed. I've added a suggestion to fix the issue.

```suggestion
	            errors.append(
	                'CourseNotFound for course "{course}". CourseEnrollment ID={rec}'.format(
	                    course=str(rec.course_id),
	                    rec=rec.id,
	                )
	            )
```

figures/backfill.py
```diff
@@ -67,7 +67,7 @@ def backfill_enrollment_data_for_site(site):
             enrollment_data.append((obj, created))
         except CourseNotFound:
             errors.append('CourseNotFound for course "{}". '
```
OmarIthawi, line 69:
```suggestion
```

johnbaldwin: @OmarIthawi Thanks, I've got a code cleanup effort to do. When I hit the pipeline logging, adding format vars and making sure all exceptions are tested will be part of that

johnbaldwin: @OmarIthawi I saw I had a missing '{}' in the string, so adding vars to the format . Thanks!

johnbaldwin:
Tox passed locally, force merging. Maybe I regret it. We'll see

PR #317, opened 2021-01-05
Release 0.4 - Tasks Error Handling and Logging Improvements

johnbaldwin:
0.4 Tasks Error Handling and Logging Improvements

1. Restructured figures.tasks daily metrics functions

The purpose of restucturing is to ensure pipeline failures do not
escalate up the call chain:

* A site failure should not impact any other site
* A course failure should not cause site processing failure
  * With the caveat that the failed course's data would not be reflected
    in the site aggregated data
* An enrollment failure should not cause the course processing to fail
  * with the caveat that the failed enrollment's data would not be
    reflected in the course aggregated data

See the module docstring in tests/tasks/test_daily_metrics.py for more
details

Logging has improvements. Namely, the logs have prefixes to help grep
through them for details. The information provided in the logs has
improved to identify the site, date for, and course id for the
processing and failures.

Logging to figures.models.PipelineError has been removed as it is not
providing benefit in its current form. Follow on work is providing post
pipeline reporting. This should provide better visibility into pipeline
data health

Restructured figures.tasks tests. Was 'tests/test_tasks.py'. Now there
are process specific modules in 'tests/tasks':

* test_daily_metrics.py
* test_monthly_metrics.py
* test_mau_tasks.py - This may go away as it was never actually used in
production. MAU is captured monthly in figures.models.SiteMonthlyMetrics
and MAU 2G (second generation) is going to be finally implemented soon
we hope

Minor fix - pipeline course daily metrics now cases 'course_id' to string when
working with figures.models.CourseDailyMetrics. This is to prevent
errors if a CourseKey is used instead of the string representation of
the course id. (redacted rant on CourseKey)

Add comments to backfill.backfill_enrollment_data_for_site

Added comments as notes to improve updating enrollment data


figures/pipeline/course_daily_metrics.py
```diff
@@ -308,7 +308,7 @@ def save_metrics(self, date_for, data):
         """
 
         cdm, created = CourseDailyMetrics.objects.update_or_create(
-            course_id=self.course_id,
+            course_id=str(self.course_id),
```
johnbaldwin, line 311: This is so it won't fail if course_id is a CourseKey

figures/pipeline/course_daily_metrics.py
```diff
@@ -339,7 +339,7 @@ def load(self, date_for=None, force_update=False, **_kwargs):
         """
         date_for = pipeline_date_for_rule(date_for)
         try:
-            cdm = CourseDailyMetrics.objects.get(course_id=self.course_id,
+            cdm = CourseDailyMetrics.objects.get(course_id=str(self.course_id),
```
johnbaldwin, line 342: This is so it won't fail if course_id is a CourseKey

devsite/devsite/settings.py
```diff
@@ -52,8 +53,6 @@
 # Set the default Site (django.contrib.sites.models.Site)
 SITE_ID = 1
 
-# TODO: Update this to allow environment variable override
-ENABLE_DEVSITE_CELERY = env('ENABLE_DEVSITE_CELERY')
```
johnbaldwin, line 56: Removed duplicate

figures/tasks.py
```diff
@@ -53,36 +58,77 @@ def populate_single_cdm(course_id, date_for=None, force_update=False):
...
 
+@shared_task
+def populate_daily_metrics_for_site(site_id, date_for, force_update=False):
+    """
+
+    """
+    try:
+        site = Site.objects.get(id=site_id)
+    except Site.DoesNotExist:
+        msg = 'add errro message'
```
johnbaldwin, line 97: Yep, I'll update to add a real error message

figures/tasks.py
```diff
@@ -95,87 +141,106 @@ def update_enrollment_data(site_id, **_kwargs):
             'figurs.tasks.update_enrollment_data: site_id={} does not exist'.format(
                 site_id))
     except Exception:  # pylint: disable=broad-except
-        msg = ('FIGURES:FAIL daily metrics:update_enrollment_data'
+        msg = ('FIGURES:DAILYLFAIL daily metrics:update_enrollment_data'
                ' for site_id={}'.format(site_id))
         logger.exception(msg)
 
 
+# TODO: Sites iterator with entry and exit logging
```
johnbaldwin, line 149: Just a note and I want to leave it here as a reminder

figures/tasks.py
```diff
@@ -95,87 +141,106 @@ def update_enrollment_data(site_id, **_kwargs):
...
-    missed when the site daily metrics record is populated.
+    This is a top level Celery task run every 24 hours to collect metrics.
 
-    NOTE: We have an experimental task that runs the course populators in
+    It iterates over each site to populate CourseDailyMetrics records for the
+    courses in each site, then populates that site's SiteDailyMetrics record.
 
-    parallel, then when they are all done, populates the site metrics. See the
-    function ``experimental_populate_daily_metrics`` docstring for details
+    Deveoper note: Errors need to be handled at each layer in the call chain
```
johnbaldwin, line 161: Typo to fix "Developer"

tests/tasks/test_daily_tasks.py
```diff
@@ -0,0 +1,282 @@
...
+    date_for = date.today()
+    site = SiteFactory()
+    fake_course_ids = ['fake-course-id-1']
+
+    def fake_pop_single_cdm_fails(**kwargs):
+        # TODO: test with different exceptions
+        # At least one with and without `message_dict`
+        raise FakeException('Hey!')
+
+    # def fake_pop_single_sdm(**_kwargs):
```
johnbaldwin, line 172: dead code I'll remove

tests/tasks/test_daily_tasks.py
```diff
@@ -0,0 +1,282 @@
...
+                    ).format(prefix=FPD_LOG_PREFIX,
+                             site_id=site.id,
+                             date_for=date_for,
+                             course_id=fake_course_ids[0],
+                             exception='Hey!')
+    assert last_log.message == expected_msg
+
+
+@pytest.mark.skipif(OPENEDX_RELEASE == GINKGO,
+                    reason='Broken test. Apparent Django 1.8 incompatibility')
```
johnbaldwin, line 195: TODO: remove 'Broken test' from reason

tests/tasks/test_monthly_tasks.py
```diff
@@ -0,0 +1,118 @@
...
+    monkeypatch.setattr('figures.tasks.populate_monthly_metrics_for_site.delay',
+                        fake_populate_monthly_metrics_for_site)
+
+    run_figures_monthly_metrics()
+
+    assert set(sites_visited) == set([rec.id for rec in expected_sites])
+
+
+@pytest.mark.skipif(OPENEDX_RELEASE == GINKGO,
+                    reason='Broken test. Apparent Django 1.8 incompatibility')
```
johnbaldwin, line 97: TODO: remove 'Broken test' from reason

figures/tasks.py
```diff
@@ -53,36 +58,77 @@ def populate_single_cdm(course_id, date_for=None, force_update=False):
...
+    SiteDailyMetricsLoader().load(site=Site.objects.get(id=site_id),
+                                  date_for=date_for,
+                                  force_update=force_update)
+
     logger.debug(
         'done running populate_site_daily_metrics for site_id={}'.format(site_id))
 
 
+@shared_task
+def populate_daily_metrics_for_site(site_id, date_for, force_update=False):
```
OmarIthawi, line 90: If I understand correctly, this is a task that we're now only calling as part of `populate_daily_metrics` but in theory we can run it independently later on. Correct?

OmarIthawi (approved):
Thanks @johnbaldwin! LGTM. I've added one question.

figures/tasks.py
johnbaldwin: Correct. I pulled it out of `populate_daily_metrics` so that we can in the future run asynchronously (using the `.delay`) but before we do, we need to make sure that RabbitMQ can handle it without blocking other async tasks

PR #320, opened 2021-01-14
porting tests from travisCI to github actions

thraxil:
Next one on our move away from TravisCI.

This ports all of the tests that we were running for Figures on Travis over to Github Actions. Kept things as line-by-line similar as possible as the first phase though I point out some areas that we could probably refactor the tests once we're comfortable with them on GHA.

The PR will not show as mergeable becuase we currently have the repository configured to expect/require passing TravisCI tests. Since this PR removes the Travis config, that obviously won't happen. Once this is approved, one of us with admin access will need to change the repo settings to instead require passing GHA tests.

The only part of the Travis config that I didn't port (for now) is the automated publishing to PyPI. I will port that over to GHA as well, but I want to do it as a separate PR once this one is in.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          pip install tox flake8
+      - name: Uses Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install -g yarn
+      - run: |
+          cd frontend && yarn install && cd ..
+      - name: flake8
```
thraxil, line 41: I copied this step over from the Travis config, but it seems like it's a bit redundant since the `lint` tox testenv runs `flake8` anyway. We probably don't need to do it separately on each.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+      - name: install dependencies
+        run: |
+          pip install tox flake8
+      - name: Uses Node.js
```
thraxil, line 34: I copied the steps that get run directly from TravisCI, so it runs `yarn` in the frontend directory before running the tox tests here just as it did on Travis. The python tests don't appear to depend on the frontend stuff in any way though. If we wanted to speed the tests up a little bit, it seems to me like we should pull out the JS stuff into its own separate test config and run that independently of the python tests.

OmarIthawi (approved):

johnbaldwin (approved):

johnbaldwin:
Thanks for doing this PR, @thraxil ! It is now out of date. Do you want me to merge master into it?

PR #321, opened 2021-01-15
get_sites: configurable backend to filter active sites

OmarIthawi:
RED-1721. This will help to make Figures use the `openedx.core.djangoapps.appsembler.sites.utils.get_active_sites` via configurations to make the pipeline run only for active sites.

This should help as both stopgap and long term solution for the increasing time of the pipeline.

### TODO

 - [x] Fix existing broken tests
 - [x] Add tests to cover both with/without backend
 - [ ] Document the setting variable?

### DIFF
--- a/figures/management/commands/backfill_figures_metrics.py
+++ b/figures/management/commands/backfill_figures_metrics.py
@@ -63,12 +63,7 @@ class Command(BaseCommand):
         if options['site']:
             sites = [get_site(options['site'])]
         else:
-            # Would be great to be able to filter out dead sites
-            # Would be really great to be able to filter out dead sites
-            # Would be really Really great to be able to filter out dead sites
-            # Would be really Really REALLY great to be able to filter out dead sites
-
-            sites = Site.objects.all()
+            sites = get_sites()
         for site in sites:
             backfill_site(site, overwrite=options['overwrite'])
 
--- a/figures/management/commands/backfill_figures_metrics.py
+++ b/figures/management/commands/backfill_figures_metrics.py
@@ -13,6 +13,7 @@ from django.core.management.base import BaseCommand
 from figures.backfill import backfill_monthly_metrics_for_site
 from figures.sites import get_sites
 
+
 def get_site(identifier):
     """Quick-n-dirty function to let the caller choose the site id or domain
     Let the 'get' fail if record can't be found from the identifier
--- a/figures/management/commands/backfill_figures_metrics.py
+++ b/figures/management/commands/backfill_figures_metrics.py
@@ -11,7 +11,7 @@ from django.contrib.sites.models import Site
 from django.core.management.base import BaseCommand
 
 from figures.backfill import backfill_monthly_metrics_for_site
-
+from figures.sites import get_sites
 
 def get_site(identifier):
     """Quick-n-dirty function to let the caller choose the site id or domain
--- a/figures/management/commands/update_figures_enrollment_data.py
+++ b/figures/management/commands/update_figures_enrollment_data.py
@@ -44,11 +44,6 @@ class Command(BaseCommand):
         if options['site']:
             sites = [get_site(options['site'])]
         else:
-            # Would be great to be able to filter out dead sites
-            # Would be really great to be able to filter out dead sites
-            # Would be really Really great to be able to filter out dead sites
-            # Would be really Really REALLY great to be able to filter out dead sites
-
             sites = get_sites()
         for site in sites:
             print('Updating EnrollmentData for site "{}"'.format(site.domain))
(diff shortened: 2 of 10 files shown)


OmarIthawi:
@johnbaldwin this isn't much ready for merge but should have everything I got in mind.

Please let me know what do you think esp. regarding the inline notes in the code.

figures/management/commands/backfill_figures_metrics.py
```diff
@@ -63,12 +64,7 @@ def handle(self, *args, **options):
         if options['site']:
             sites = [get_site(options['site'])]
         else:
-            # Would be great to be able to filter out dead sites
-            # Would be really great to be able to filter out dead sites
-            # Would be really Really great to be able to filter out dead sites
-            # Would be really Really REALLY great to be able to filter out dead sites
-
-            sites = Site.objects.all()
+            sites = get_sites()
```
johnbaldwin, line 67: Yay!

figures/management/commands/update_figures_enrollment_data.py
```diff
@@ -43,12 +44,7 @@ def handle(self, *args, **options):
         if options['site']:
             sites = [get_site(options['site'])]
         else:
-            # Would be great to be able to filter out dead sites
-            # Would be really great to be able to filter out dead sites
-            # Would be really Really great to be able to filter out dead sites
-            # Would be really Really REALLY great to be able to filter out dead sites
-
-            sites = Site.objects.all()
+            sites = get_sites()
```
johnbaldwin, line 47: Yay!

figures/sites.py
```diff
@@ -289,3 +289,31 @@ def site_certificates(site):
...
+def get_sites():
+    """
+    Get a list of sites for Figures purposes in a configurable manner.
+
+    This functions makes use of the `SITES_BACKEND` setting if configured, otherwise
+    it defaults to  _get_all_sites().
+
+    :return list of Site (QuerySet)
+    """
+    sites_backend_path = settings.ENV_TOKENS['FIGURES'].get('SITES_BACKEND')
```
johnbaldwin, line 312: Just to note: Confirming it is ok to assume `settings.ENV_TOKENS['FIGURES']` will always exist.

Figures sets a default empty dict for `settings.ENV_TOKENS['FIGURES']`. See here: https://github.com/appsembler/figures/blob/0.4.dev7/figures/settings/lms_production.py#L80

figures/sites.py
```diff
@@ -289,3 +289,31 @@ def site_certificates(site):
...
+    if sites_backend_path:
+        sites_backend = import_from_path(sites_backend_path)
+        sites = sites_backend()
```
johnbaldwin, line 315: Was first thinking, why not just do `sites = import_from_path(sites_backend_path)()` But this is probably clearer

figures/tasks.py
```diff
@@ -378,8 +376,8 @@ def populate_monthly_metrics_for_site(site_id):
 @shared_task
 def run_figures_monthly_metrics():
     """
-    TODO: only run for active sites. Requires knowing which sites we can skip
+    Populate monthly metrics for all sites.
```
johnbaldwin, line 379: Technically, this is is only *all* sites when a filtered sites backend is not used

figures/views.py
```diff
@@ -899,8 +899,10 @@ class SiteViewSet(StaffUserOnDefaultSiteAuthMixin, viewsets.ReadOnlyModelViewSet
     Access is restricted to global (Django instance) staff
     """
     model = Site
-    queryset = Site.objects.all()
```
johnbaldwin, line 902: You should probably just be able to set `queryset = figures.sites.get_sites()` and not need the `get_queryset` method. I'd need to look through the DRF code to see if there is any significant difference

tests/test_helpers.py
```diff
@@ -220,3 +221,30 @@ def test_first_last_days_for_month():
     assert last_day.year == year
     assert first_day.day == 1
     assert last_day.day == 29
+
+
```
johnbaldwin, line 225: @OmarIthawi Nice simple set of tests here, I like them!

figures/tasks.py
OmarIthawi, line 379:
```suggestion
    Populate monthly metrics for sites.
```

figures/views.py
OmarIthawi: If `get_sites` returns an already [evaluated list (and memory cached)](https://stackoverflow.com/a/40618173/161278) there could be a problem of having the list of sites stale. 

If it did have a stale sites, it means the only way to refresh the list is to restart the workers.

While this issue may not happen, it would be a rabbit hole to discover it.

Therefore I'm erring on the safe side of using a method since it's out of the control of Figures.

OmarIthawi:
@johnbaldwin where to document the Figures settings? If there's no doc, where should put it?

figures/views.py
johnbaldwin: Fair enough

johnbaldwin (approved):
LGTM, thanks @OmarIthawi !

PR #327, opened 2021-03-17
Create PULL_REQUEST_TEMPLATE

ahmedaljazzar:
Encouraging contributors to write documentation and tests for their PRs whenever needed. I found it really helpful in Wharf repo, and we are applying it in all major repos.

### DIFF
--- a/PULL_REQUEST_TEMPLATE
+++ b/PULL_REQUEST_TEMPLATE
@@ -1,0 +1,14 @@
+### Overview
+
+- What sparked this PR? 
+- What problem does it solve? 
+- Why is it useful?
+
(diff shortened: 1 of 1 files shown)


johnbaldwin:
@ahmedaljazzar Thanks for the idea. Just letting you know I see your PR here. I'll leave this one open until I create an issue to shape the broader context. We have internal contributors, an external contributor and if things go in the direction we think, there will be more external contributors. Therefore, we have to make the message very clear (doesn't have to be long, short is good), but we have to make the message clear

OmarIthawi:
Closed in favor of #366

PR #334, opened 2021-04-29
experiment with waffle swtich to disable tasks for maintenance

OmarIthawi:


OmarIthawi:
@johnbaldwin That's an experiment with Waffle switches.

I works well on one task (minus the `django-waffle` pip package).

However, it has to be added for each task invividually.

Our options are:

 - Add to each task, with tests (bad: test noise)
 - Make a function decorator `@switchable_task` which would decorate each task (bad: decorators can be complex)
 - Disable the pipeline via settings `ENABLE_FIGURES_MONTHLY_METRICS: false` etc (bad: require a deployment before the cutover)

What do you think?

thraxil (approved):

figures/tasks.py
```diff
@@ -42,6 +43,7 @@
 # TODO: Make this configurable in the settings
 # logger.setLevel('INFO')
 
+WAFFLE_DISABLE_PIPELINE = 'figures.disable_pipeline'
 
 @shared_task
```
thraxil, line 48:
```suggestion

@shared_task
```

figures/tasks.py
```diff
@@ -50,6 +52,10 @@ def populate_single_cdm(course_id, date_for=None, force_update=False):
     The calling function is responsible for error handling calls to this
     function
     """
+    if waffle.switch_is_active(WAFFLE_DISABLE_PIPELINE):
+        logger.info('%s: Disabled due to %s being active.', populate_single_cdm.__name__, WAFFLE_DISABLE_PIPELINE)
```
thraxil, line 56:
```suggestion
        logger.info('%s: Disabled due to %s being active.', populate_single_cdm.__name__,
            WAFFLE_DISABLE_PIPELINE)
```

thraxil:
Just a couple flake8 fixes.

figures/tasks.py
```diff
@@ -50,6 +53,11 @@ def populate_single_cdm(course_id, date_for=None, force_update=False):
     The calling function is responsible for error handling calls to this
     function
     """
+    if waffle.switch_is_active(WAFFLE_DISABLE_PIPELINE):
```
johnbaldwin, line 56: @OmarIthawi Thanks for trying, but this will not work.

We can't just put a waffle flag in only the CDM task as the SDMs will get still created and LearnerMetrics records will get created/updated, but the data will be wrong, so we'd have to do backfill data repair _anyway_. The place for 
We're going to need the flag to be in the 

Please read my comment here: https://appsembler.atlassian.net/browse/RED-2075?focusedCommentId=34163

If we are waffle flagging tasks, (instead of settings, where celerybeat jobs are assigned), then the places to do it are in `figures.sites.populate_daily_metrics` and `figures.sites.run_figures_monthly_metrics`

johnbaldwin (changes requested):
If we're waffle flagging figures.tasks, the functions that need the flag are:
* figures.sites.populate_daily_metrics 
* figures.sites.run_figures_monthly_metrics

OmarIthawi:
Thanks @johnbaldwin

> We can't just put a waffle flag in only the CDM task as the SDMs will get still created and LearnerMetrics records will get created/updated, but the data will be wrong, so we'd have to do backfill data repair _anyway_. The place for 
> We're going to need the flag to be in the 
> Please read my comment here: https://appsembler.atlassian.net/browse/RED-2075?focusedCommentId=34163

👍🏼 

> If we are waffle flagging tasks, (instead of settings, where celerybeat jobs are assigned), then the places to do it are in `figures.sites.populate_daily_metrics` and `figures.sites.run_figures_monthly_metrics`
> 

> If we're waffle flagging figures.tasks, the functions that need the flag are:
> 
> * figures.sites.populate_daily_metrics
> * figures.sites.run_figures_monthly_metrics

Thanks @johnbaldwin. That helps. Do you recommend using Waffle Flags? or should we use django settings?

johnbaldwin:
@OmarIthawi I haven't yet played with waffle flags, bu AFIAK, Waffle flags are toggled at runtime. Django settings are not, we've got to update the settings (PR review on the configs), then push the update. Being able to turn the pipeline trigger on and off with a runtime switch is a useful feature. It solves the planned maintenance window use case. 

NOT NEEDED NOW, maybe premature optimization: A bonus is to waffle flag per site with the default to be on. This way if we want, we can disable processing for specific active sites

OmarIthawi:
> @OmarIthawi I haven't yet played with waffle flags, bu AFIAK, Waffle flags are toggled at runtime. Django settings are not, we've got to update the settings (PR review on the configs), then push the update. Being able to turn the pipeline trigger on and off with a runtime switch is a useful feature. It solves the planned maintenance window use case.

Thanks for checking that, I'll get to it with Esther soon.

> NOT NEEDED NOW, maybe premature optimization: A bonus is to waffle flag per site with the default to be on. This way if we want, we can disable processing for specific active sites

👍🏼

OmarIthawi:
I'll close in favor of https://github.com/appsembler/figures/pull/335

PR #345, opened 2021-05-19
Reorganize the requirements files for Juniper tox tests

johnbaldwin:
The purpose of this commit is to clean up the requirements files. This makes things a bit clearer. We now have a "Juniper community" mode and a "Juniper multisite" mode.

* We remove Py38 because Juniper needs to run in Py35
* The `juniper_community.txt` and `juniper_multisite.txt` requirements files read from `juniper_base.txt`. Each of the 'community' and 'multisite' have only the package differences, which at this point is the version/branch of `edx-organizations`

This PR is in preparation for another PR to improve asset building and testing devsite to support S3 based assets as well as local based assets

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,9 +15,9 @@ jobs:
           - python: 2.7
             tox-env: py27-hawthorn_multisite
           - python: 3.5
-            tox-env: py35-juniper
-          - python: 3.8
-            tox-env: py38-juniper
+            tox-env: py35-juniper_community
+          - python: 3.5
+            tox-env: py35-juniper_multisite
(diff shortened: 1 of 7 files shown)


thraxil (approved):

johnbaldwin:
@thraxil Do you have any idea where `test (3.5, py35-juniper)` and `test (3.8, py38-juniper) ` are coming from? I've grepd and looked over manuall the .github files and tox.ini but can't seem to locate these. I don't see them in other open PRs. Could this be an artifact of me changing the .github/workflows/test.yml file after initially creating the PR?

OmarIthawi (approved):

PR #351, opened 2021-05-27
Handle old, slash-separated course ids and course-vX:... style

bryanlandia:
Allows Figures to run on sites with courses created in e.g., 2015

Split route path into two options:  if there are no forward-slashes we assume course id is like course-v1:xxx 
I didn't figure out how to get a nice named match param into my regex for that one but if you see how to, feel free.
Second one looks specifically for a course id style with two slash separations.


johnbaldwin (approved):
Looks reasonable to me. Too bad we don't have unit testing on the front end. Are you able to get this to work in Figures devsite @bryanlandia or did you test it elsewhere?

I would like to get @grozdanowski OK on this before we merge it in case he sees something I don't

bryanlandia:
Thanks @johnbaldwin I have just been testing directly against DHIS2 staging :D And yes, it's working for old and new style course ids.

grozdanowski (approved):
Thanks for this, @bryanlandia

bryanlandia:
Thanks guys. I'll rebase and merge

PR #352, opened 2021-05-28
Refactor and fix Backfill commands 

bryanlandia:
This PR addresses some shortcomings of previous Figures data backfill picture.  It does some things:

# The old stuff:

Marks backfill_figures_metrics and populate_figures_metrics commands as pending deprecation.  Replaces their handlers with calls to the new backfill commands.

# The new stuff:

## BaseBackfillCommand

You can pass a start date, end date, site id/domain, overwrite yes/no, and no_delay yes/no to all commands below subclassing this. Ovewrite will replace existing objects, no_delay will directly call Celery tasks instead of queueing them via RabbitMQ

## backfill_figures_daily_metrics

* Creates all CourseDailyMetrics records for a date range and site (or all sites if none passed).  
* Creates all SiteDailyMetrics records for a date range and site (or all sites if none passed).  
* Refuses to calculate average_progress for a date previous to yesterday, since this will never be accurate.  The added benefit is this removes the major cause of slow processing (calculting grades).

Uses the experimental method if that option is passed, but I didn't worry much about that, since I don't think it was every really finished.

## repair_figures_backfilled_progress

* Sets average_progerss to None for dates prior yesterday within a passed date range, with a dry run option.  
* Used to correct prior backfills which would have set all dates' average_progress to the current day's average_progress for that course

## backfill_figures_monthly_metrics

* Creates all SiteMonthlyMetrics for a date range and site (or all sites if none passed)

## backfill_figures_enrollment_data

* Creates EnrollmentData objects for each course with enrollments for each site passed, or all sites if no site passed


bryanlandia:
@johnbaldwin I addressed your comment (thanks!)  and this is ready for your review now.

figures/management/commands/__init__.py
```diff
@@ -0,0 +1,74 @@
+"""
```
johnbaldwin, line 1: @bryanlandia I'm not a fan of actually adding any real code to `__init__.py` modules unless there is some super compelling reason. I'd much rather use a "named" module. I understand in this case any non `__init__.py` files in 'commands' directory implies it is a command, which is a bit of a pain. So I'm not going to be dogmatic about it. I am curious what you think about creating a `management/base.py` module, but won't ask for this for this PR. We can improve this in the future.

johnbaldwin:
@bryanlandia There's a bunch here and it is way too late for me, so I'm going to pick up the review again tomorrow to make sure I don't miss something. My biggest concern is that we don't break existing django management commands that would be used with cron instead of celerybeat, namely that this PR removes `populate_figures_metrics.py` I'd rather we keep this command for anyone in the community who might be using it. I'm happy with making deprecation notes, but would prefer we keep this command

figures/management/commands/__init__.py
bryanlandia: Interesting.  Yes, I think management/base.py is a good idea.

johnbaldwin:
@bryanlandia I've looked over the backfill commands. Aside from preserving `populate_figures_metrics` as mentioned above, I think it best to have `figures` in the backfill command names. For example: `backfill_figures_monthly_metrics`.

I ask this only as a measure against management command naming collisions. The chance of this is pretty low now. I don't see any other commands with `backfill` except for `backfill_history` for the `course_overviews` app. But this just helps avoid that.

The only other interface related thing with backfill commands is that I'm interested in what you see is the advantage of having multiple different commands rather than have one `backfill_figures` command that takes the first positional argument as a choice, like `./manage.py lms backfill_figures monthly ...`. Definitely not something critical now, but thinking longer term

johnbaldwin (approved):

figures/management/base.py
```diff
@@ -0,0 +1,75 @@
+"""
+Management command base classes for Figures.
+"""
+from datetime import datetime
+
```
estherjsuh, line 5: import argparse

figures/management/commands/populate_figures_metrics.py
```diff
@@ -48,39 +48,31 @@ def add_arguments(self, parser):
...
-                    experimental_populate_daily_metrics.delay(**kwargs)  # pragma: no cover
-            else:
-                if options['no_delay']:
-                    populate_daily_metrics(**kwargs)
-                else:
-                    populate_daily_metrics.delay(**kwargs)  # pragma: no cover
+            call_command(
+                'backfill_figures_daily_metrics',
+                no_delay=options['no_delay'],
+                date_start=options['date'],
```
estherjsuh, line 71: @bryanlandia 
should this be `options['date_start']` ?

I'm not super familiar with argument parsers but the objects we created so far are: date_start, date_end, experimental, site, overwrite, and no_delay.

figures/management/commands/populate_figures_metrics.py
```diff
@@ -48,39 +48,31 @@ def add_arguments(self, parser):
...
+                date_end=options['date'],
```
estherjsuh, line 72: @bryanlandia and this one `options['date_end']` ?

figures/management/commands/populate_figures_metrics.py
bryanlandia: Thanks for taking a look!  
In this case,  options['date'] is the argument passed to populate_figures_metrics, but the new command we are calling wants  arguments called date_start and date_end, so we just pass `options['date']` to it as the value for date_start *and* date_end.

figures/management/commands/populate_figures_metrics.py
bryanlandia: ditto :)

estherjsuh (approved):
looks good! 🥇

PR #353, opened 2021-05-29
Juniper Upgrade - Fix the front end asset build

johnbaldwin:
There were a couple of problems

1. The multi-build.js would not call the second call in `require('./build.js')
2. Assets were not being created in the path specified by the 'FIGURES_APP_BUILD' directory

This PR fixes those and sets the default build target, `npm run build` to build both sets of assets for each of the REM_BASE=10 and REM_BASE=16 environments.


melvinsoft (approved):

OmarIthawi (approved):

bryanlandia:
Awesome to start work today and see this fixed!  Was running into an issue last Thu trying to deploy a temp branch to a Hawthorn server

Issue #356, opened 2021-05-31
February is skipped in monthly metrics histories


bryanlandia:
Appears that we never get any kind of metrics for February (unless probably it's the current month).  This affects all types of metrics.

<img width="1378" alt="image" src="https://user-images.githubusercontent.com/1289191/120234842-a1cf1a80-c20d-11eb-9f94-7fdfb353451b.png">

Looks like the problem is in `metrics.previous_months_iterator`

```python
>>> from datetime import datetime
>>> date_for = datetime.utcnow().date()
>>> from figures import helpers, metrics
>>> date_for = helpers.as_date(date_for)
>>> date_for
datetime.date(2021, 5, 31)
>>> metrics.previous_months_iterator(month_for=date_for, months_back=6)
<generator object previous_months_iterator at 0x7fbc30ec0190>
>>> [mon for mon in metrics.previous_months_iterator(month_for=date_for, months_back=6)]
[(2020, 11, 30), (2020, 12, 31), (2021, 1, 31), (2021, 3, 31), (2021, 4, 30), (2021, 5, 31)]
```

bryanlandia:
I haven't yet verified if Feb's data is appearing as January or if February is just skipped altogether.

bryanlandia:
See note here https://dateutil.readthedocs.io/en/stable/rrule.html#classes  

> Per RFC section 3.3.10, recurrence instances falling on invalid dates and times are ignored rather than coerced:

> Recurrence rules may generate recurrence instances with an invalid date (e.g., February 30) or nonexistent local time (e.g., 1:30 AM on a day where the local time is moved forward by an hour at 1:00 AM). Such recurrence instances MUST be ignored and MUST NOT be counted as part of the recurrence set.

> This can lead to possibly surprising behavior when, for example, the start date occurs at the end of the month:

>>> from dateutil.rrule import rrule, MONTHLY
>>> from datetime import datetime
>>> start_date = datetime(2014, 12, 31)
>>> list(rrule(freq=MONTHLY, count=4, dtstart=start_date))
... # doctest: +NORMALIZE_WHITESPACE
[datetime.datetime(2014, 12, 31, 0, 0),
 datetime.datetime(2015, 1, 31, 0, 0),
 datetime.datetime(2015, 3, 31, 0, 0),
 datetime.datetime(2015, 5, 31, 0, 0)]

bryanlandia:
See draft PR #358 for fix

johnbaldwin:
[...]

PR #358, opened 2021-05-31
Fix previous_months_iterator

bryanlandia:
Addresses #356

Avoid rrule when possibly iterating forward from days 29-31 of months,
rrule simply skip month if it tries to iterate to e.g., February 30.


bryanlandia:
Hi @johnbaldwin I've updated the helper method and associated test method.  The previous_months_iterator is now working properly across short months, leap years, and returns the correct number of months back.

johnbaldwin (approved):
@bryanlandia Looks reasonable to me. Thanks!

Issue #361, opened 2021-06-03
Figures' /courses/... URL route matches request_utils' COURSE_REGEX, causing remainder to be interpreted as a course id


bryanlandia:
I am seeing 500 errors calling /courses/general/course-v1:... for the courses-general API

The API call isn't even getting processed by the DRF view, as it errors out in middlewares which interpret `/general/course-v1:...` as a course id.

For example,  this traceback:

```
2021-06-03 17:13:22,487 WARNING 7580 [track.contexts] contexts.py:20 - unable to parse course_id "general/course-v1:HISP+D2EVENTS100+Q2_2020"
Traceback (most recent call last):
  File "/edx/app/edxapp/edx-platform/common/djangoapps/track/contexts.py", line 20, in course_context_from_url
    url = url or ''
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/opaque_keys/__init__.py", line 197, in from_string
    return cls.deprecated_fallback._from_deprecated_string(serialized)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/opaque_keys/edx/locator.py", line 379, in _from_deprecated_string
    raise InvalidKeyError(cls, serialized)
InvalidKeyError: <class 'opaque_keys.edx.locator.CourseLocator'>: general/course-v1:HISP+D2EVENTS100+Q2_2020
2021-06-03 17:13:22,777 ERROR 7580 [root] signals.py:17 - Uncaught exception from None
Traceback (most recent call last):
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 244, in _legacy_get_response
    response = middleware_method(request)
  File "/edx/app/edxapp/venvs/edxapp/src/organizations/organizations/middleware.py", line 32, in process_request
    org = CourseKey.from_string(course_id).org
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/opaque_keys/__init__.py", line 197, in from_string
    return cls.deprecated_fallback._from_deprecated_string(serialized)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/opaque_keys/edx/locator.py", line 379, in _from_deprecated_string
[...]

PR #362, opened 2021-06-03
Fix frontend API url courses regex overlap

bryanlandia:
Addresses #361 by changing 

* /figures/api/courses/general to /figures/api/courses-general/
* /figures/api/courses/detail to /figures/api/courses-detail/

Also, just to not have one outlier URL pattern, updates the corresponding 

* /figures/api/users/general to /figures/api/users-general/
* figures/api/users/detail to /figures/api/users-detail/

These changes shouldn't be problematic in that they are internal API URLs and the documentation has warnings that they are subject to change.


bryanlandia:
Looks like failing test is unrelated, re: #360

johnbaldwin (approved):
@bryanlandia This PR has outdated code. Please make sure this passes tests when you merge it back in. I'm going to consider this a "temporary" workaround fix. Why? Because nested resources are valid REST endpoints. Too much nesting is a code smell, but I don't want to box figures out of being able to do endpoints in the future for:
```
/courses
/courses/some-blah
/courses/some-other-blah
```

bryanlandia:
@johnbaldwin thanks for the review!  Unfortunately because the course id regex is defined in edx-platform code, and evaluated in middleware, any `/courses/blah` URL pattern is going to run into trouble.  I'll merge after rebase as long as all the checks pass.

PR #366, opened 2021-06-07
Add Vanta required PR and Issue templates

thraxil:
All repos must have PR and Issue templates now. These aren't perfect, but are something to start with.


johnbaldwin (approved):
@thraxil So, I'm not thrilled with this, I find this one of the "straws on the camel's back" type nuisances, I'll approve it since I understand the necessity to have something for compliance. What would be great is to have some documentation link to explain what all this is about for context for the community. Without context these kinds of forced processes scream process with a big "P" for the sake of process rather that process with a little "p" for the sake of making the product better and safer

thraxil:
> Without context these kinds of forced processes scream process with a big "P" for the sake of process rather that process with a little "p" for the sake of making the product better and safer

Welcome to SOC2.

PR #382, opened 2021-06-22
Allow site arg by domain name again for backfill_figures_daily_metrics

bryanlandia:
My last PR wasn't quite right, since it did away with the option to pass a domain name instead of Site id—way more convenient. This brings back that ability for backfill_figures_daily_metrics

### DIFF
--- a/figures/management/commands/backfill_figures_daily_metrics.py
+++ b/figures/management/commands/backfill_figures_daily_metrics.py
@@ -51,13 +51,16 @@ class Command(BaseBackfillCommand):
             date_start, date_end
         ))
 
+        # don't pass multiple site ids to tasks
+        site_id = None if not options['site'] else self.get_site_ids(options['site'])
+
         # populate daily metrics one day at a time for date range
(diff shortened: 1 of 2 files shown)


figures/management/commands/backfill_figures_daily_metrics.py
```diff
@@ -51,13 +51,16 @@ def handle(self, *args, **options):
             date_start, date_end
         ))
 
+        # don't pass multiple site ids to tasks
+        site_id = None if not options['site'] else self.get_site_ids(options['site'])
```
estherjsuh, line 55: @bryanlandia I was looking at the docstring for `get_site_ids`.
If `site_id` is None, `get_site_ids` method stores returns all available sites ids as a list. Would this be passing multiple site ids to task?

        """Quick-n-dirty function to let the caller choose the site id or domain.
        If no identifier is passed, return all available Sites.
        Let the 'get' fail if record can't be found from the identifier.
        Returns Site ids for passing to Celery tasks.
        Note that at present, none of the tasks handle more than one specified Site.
        """

estherjsuh (approved):

PR #387, opened 2021-07-08
fix flaky `test_backfill_monthly_metrics_for_site` by freezing time

OmarIthawi:
`test_backfill_monthly_metrics_for_site` seems to fail sometimes depending on the time it's being run. using [pytest-freezegun](https://pypi.org/project/pytest-freezegun/) to freeze the time and make sure it returns consistent results.

### Old failure on Jun 30, 2021

image

### DIFF
--- a/devsite/requirements/test.txt
+++ b/devsite/requirements/test.txt
@@ -1,0 +1,2 @@
+# Shared testing requirements
+pytest-freezegun==0.4.2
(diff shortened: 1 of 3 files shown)


thraxil (approved):
Yes, we meet again `datetime.now()`...

OmarIthawi:
> Yes, we meet again `datetime.now()`...

I knew you'd be "excited"

PR #401, opened 2021-11-15
add bandit security tests

thraxil:
Bandit finds a couple issues (see `tox -e bandit` for the report). Those should be addressed separately. Once everything is passing, we can add it to the Github Actions workflow.


johnbaldwin (approved):
Adding doc link for context: https://bandit.readthedocs.io/en/latest/

PR #414, opened 2022-01-26
Fixes serializer tests failing locally in Ginkgo

johnbaldwin:
* The version 2.4.2 `dateutil.parser.parse` method returns a `datetime`
object in local time, causing the tests to fail when compared against
datetime in UTC
* This commit addresses the issue by replacing the timezone with UTC
* This commit includes a helper function, `as_datetime_utc` in tests/test_serializers.py
* Read the docstring for this helper function

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -3,8 +3,11 @@
 
 from __future__ import absolute_import
 import os
+from dateutil.parser import parse as dateutil_parse
 from dateutil.rrule import rrule, DAILY
 from packaging import version
+
+from django.utils.timezone import utc
 
 from organizations.models import Organization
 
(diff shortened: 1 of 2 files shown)


estherjsuh (approved):

bryanlandia (approved):
Makes sense, but would this function be better put into a test utils module rather than just in this specific serializers test module?  

https://github.com/appsembler/figures/search?q=dateutil.parser

shows a lot of files using a `dateutil_parse`.  Maybe at least put this func in test_helpers.py

I'll approve but please consider

johnbaldwin:
@bryanlandia I updated per your request

bryanlandia:
Thanks @johnbaldwin Looks like as_datetime_utc could probably be used here too 
https://github.com/appsembler/figures/blob/95ac4ac2153166387a7142f12a92908efa8c0647/tests/test_filters.py#L207

PR #427, opened 2022-02-24
Pipeline improvement prerequisites

johnbaldwin:
This PR contains updates to Figures as prerequisites to adding the second switchable workflow

Each PR has it's own description and context. So please read the changes in calendar order so you see the details to the story.

## Why this PR exists and high level story

1. We need to improve Figures daily metrics collection job  (the "daily pipeline") performance for one of our standalone deployments
2. The existing daily job is working on Tahoe and we don't want to break it
3. So we add a second daily job workflow
4. We enable the deployment to choose which workflow to run on a setting in Figures server-vars
5. After we prove the robustness of the second workflow, we will deprecate and then remove the original workflow

## What's the scope of this PR

1. This PR contains code refactoring and changes needed to support the new workflow
2. This PR does NOT contain the actual second workflow code
3. These are broken up into two PRs for readability

## What's IN this PR?

1. Added a Figures `ENV_TOKENS` flag to choose an override task to kick of the daily Celery top level task or use the default one: https://github.com/appsembler/figures/commit/926cd7c5bb22dd1670ee221de473c071e4aad1d8
2. Minor code cleanup found along the way that is opportune to include in this PR: https://github.com/appsembler/figures/commit/83687ca54b77d90dfaafaeb2a2301ea352ece5f6
3. Because it can be confusing to get the specific calendar day (24 hour period starting at 00:00 UTC) for "yesterdays" data collection, added a convenience function to `figures.helpers`: https://github.com/appsembler/figures/commit/42ce02d575ba9cc6004d4dde91699252daf9f7f0
4. Found a bug in Figures devsite settings when I needed to create a migration, so fixed that here: https://github.com/appsembler/figures/commit/5a5ff31e03ed6a6ca18e9502effcfb589cd78b13
5. And the most important part of this PR, updated metrics collection handling to encapsulate and abstract "per enrollment" progress collection here: https://github.com/appsembler/figures/commit/c6693fe12d7873dd94017c438c5ac32cd427f953

It's the last PR that is most important (I repeat myself because it is the most important. There, I've done it again). Anyway, with the new workflow, we will invert the relationship of `LearnerCourseGradeMetrics` (LCGM) and `EnrollmentData` where we write first to `EnrollmentData` (let's call him "ed") and use ed as the primary for data retrieval in the API (we started that already when we introduced ed as a model in figures). LCGM gets pushed back to be a historical record to track the time series per-enrollment performance

But we're not there yet, lots of refactoring to do to get there. What no. 5 commit above does is abstract this relationship from the rest of Figures and isolate the progress retrieval to make Figures safer, faster and more fun to program on

### DIFF
--- a/figures/helpers.py
+++ b/figures/helpers.py
@@ -1,6 +1,7 @@
 """Helper functions to make data handling and conversions easier
 
 # Figures 0.3.13 - Defining scope of this module
+# Figures 0.4.x - Yep, this module is still handy and the scope hasn't exploded
 
 The purpose of this module is to provide conveniece methods around commonly
 executed statements. These convenience methods should serve as shorthand for
@@ -34,6 +35,8 @@ repeat yourself) the code and make the code more readable
 
 ## What does not belong here?
 
+* Most importantly, if you have to import from another figures module, it does
+  not belong here!
 * "Feature" functionality does not belong here
 * Long functions do not belong here
 * Code that communicates outside of Figures does not belong here. No database,
@@ -42,7 +45,10 @@ repeat yourself) the code and make the code more readable
 This is not an exhaustive list. We'll grow it as needed.
 
 An important point is that we should not expect this module to be a permanent
-home for functionality.
+home for the functionality included. As we evolve Figures, we may find functions
+here that have a stronger context with another module. For example, we've got
+a decent set of date oriented functions that are candidatdes for a datetime and
+date handling module.
 """
 
 from __future__ import absolute_import
@@ -186,6 +192,15 @@ def next_day(val):
 
 def prev_day(val):
     return days_from(val, -1)
+
+
+def utc_yesterday():
+    """Get "yesterday" form the utc datetime
+
+    We primarily use this for the daily metrics collection. However, it proves
+    handy as a convenience function for exploring data in the Django shell
+    """
+    return prev_day(datetime.datetime.utcnow().date())
 
 
 def days_in_month(month_for):
--- a/figures/models.py
+++ b/figures/models.py
@@ -5,7 +5,7 @@ TODO: Create a base "SiteModel" or a "SiteModelMixin"
 
 from __future__ import absolute_import
 from datetime import date
-import timeit
+from time import time
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.validators import MaxValueValidator, MinValueValidator
@@ -283,7 +283,7 @@ class EnrollmentDataManager(models.Manager):
 
         if not ed_recs or ed_recs[0].date_for < date_for or force_update:
             # We do the update
-            start_time = timeit.default_timer()
+            start_time = time()
             # get the progress data
             ep = EnrollmentProgress(user=course_enrollment.user,
                                     course_id=str(course_enrollment.course_id))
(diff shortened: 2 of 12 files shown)


OmarIthawi:
Thanks John! I'll review it today.

figures/helpers.py
```diff
@@ -188,6 +194,15 @@ def prev_day(val):
     return days_from(val, -1)
 
 
+def utc_yesterday():
+    """Get "yesterday" form the utc datetime
+
+    We primarily use this for the daily metrics collection. However, it proves
+    handy as a convenience function for exploring data in the Django shell
+    """
+    return prev_day(datetime.datetime.utcnow().date())
```
thraxil, line 203: My standard advice anytime I see `datetime.now()` or an equivalent, is that you should allow a `now` parameter to be passed in in place of it. Ie,

```python
def utc_yesterday(now=None):
    if not now:
        now = datetime.datetime.utcnow()
    return prev_day(now.date())
```

This makes it considerably easier to test, particularly for the weird edge cases that come up around dates (daylight savings, leap years, etc.) and eliminate the flakey tests that can come up when tests potentially behave differently depending on when they are run (what happens when this code is run exactly at midnight?).

Code that calls `utc_yesterday()` also ought to allow their "now" to be overridden for the same reasons, though they would then only need to pass that along to `utc_yesterday()`. That also tends to make it simpler to implement things like backfill operations, since you can run some code as if it were run at some different time.

figures/models.py
```diff
@@ -5,6 +5,7 @@
 
 from __future__ import absolute_import
 from datetime import date
+import timeit
```
OmarIthawi, line 8: Do you `timeit` safe for production use? I don't know. I've checked https://docs.python.org/3/library/timeit.html and it shows a GCP related note:

> Note By default, [timeit()](https://docs.python.org/3/library/timeit.html#timeit.Timer.timeit) temporarily turns off [garbage collection](https://docs.python.org/3/glossary.html#term-garbage-collection) during the timing. The advantage of this approach is that it makes independent timings more comparable. The disadvantage is that GC may be an important component of the performance of the function being measured. If so, GC can be re-enabled as the first statement in the setup string. For example

I think it's only related to the following usage `timeit.Timer('for i in range(10): oct(i)', 'gc.enable()').timeit()`.

I think it's safe, but want to double check if we've missed something here.

figures/helpers.py
```diff
@@ -188,6 +194,15 @@ def prev_day(val):
     return days_from(val, -1)
 
 
+def utc_yesterday():
```
OmarIthawi, line 197: Would you mind adding a test case for this function? One happy scenario should be enough.

OmarIthawi (approved):
Thanks @johnbaldwin! Looks good to me.

I've added one question about `timeit` and tests for `utc_yesterday` but none are blockers.

figures/helpers.py
johnbaldwin: You raise an excellent point make a good argument. I started to draft a reply, but then the reply started to grow longer. So I moved that off for later follow up so I can focus on getting part 2 in PR and come back to this topic. I just wanted to thank you now for reviewing so soon after I posted the PR before the weekend

figures/models.py
johnbaldwin: Thank you Omar! To be on the safe side, I'll just capture start and end times without timeit

figures/helpers.py
johnbaldwin: will do

Issue #428, opened 2022-02-25
Update enrollment data handling - "Enrollment Data Next"


johnbaldwin:
The purpose of this issue is to briefly describe an identified performance issue with collecting per-enrollment data and a high level of Figures next iteration for handling per-enrollment data collection and management.

## The problem

Figures daily metrics job gets slower as the number of `StudentModule` records gets bigger. In the daily job, Figures checks `StudentModule` records for any changes since the last time we saved a `LearnerCourseGradeMetrics` record for the enrollment. We check the `modified` data of the latest `StudentModule` for the enrollment against the latest `date_for` of the enrollment's `LearnerCourseGradeMetrics` records. If the `StudentModule` is newer, we update Figures data store for the enrollment. _Aside: I'm sorry, that model name is awful and it is my fault and I will eventually make this right_. 

In brief, Figures does this on each enrollment in a course. It's a simple architecture, has worked really well. However, while this woks fine for small data sets or for beefy databases, it runs into performance issues with larger data sets and less beefy databases. Why? Because MySQL needs to do dynamic sorting on `StudentModule` on `modified` with filtering on the user id and course id. And `edx-platform` appears to not have indexed this.

We need a solution "like yesterday" and we want a portable solution for the community too. So that ruled out adding indexes to `StudentModule`. There may be a way to speed up the queries with raw SQL, but we'd prefer to keep with vanilla Django queryset API.

## The solution we implement
We are implementing a second workflow that can be enabled by setting a flag in Figures `ENV_TOKEN` settings. This flag allows existing deployments to continue to use the current workflow while we deploy and validate correctness and performance of the new workflow in a production environment with the same Figures release. After we've validated it, we will then make the new workflow the default, deprecating the original, then remove the original workflow from Figures codebase. So, what's the new workflow?

## The new workflow, aka "Enrollment Data Next"

* First, for each course we process, we query for `StudentModule` records created or modified on the previous calendar day. This is the twenty four hour period starting at 00:00 UTC. Since we don't need to sort them to find the latest, this is a pretty fast query. We then get the distinct enrollments matching these `StudentModule` records and iterate over each of these `CourseEnrollment` records. This removes the need to query `StudentMdule` for each enrollment as we iterate over enrollments in the course to collect the previous day's metrics.
* Second, we can then simply query `CourseGradeFactory()` to retrieve the section/subsection tree structure and collect the progress data

## See related PRs

* https://github.com/appsembler/figures/pull/427
[...]

PR #429, opened 2022-02-25
Figures pipeline performance improvement

johnbaldwin:
This PR contains code to improve Figures pipeline performance for collecting daily metrics

This PR follows on https://github.com/appsembler/figures/pull/427

## What is in this PR?

* `figures.course` - A new module to simplify retrieving data about a specific course. See commit, https://github.com/appsembler/figures/pull/429/commits/839b3226933215c9e4b4e6233ab941707498bc00
* `figures.pipeline.enrollment_metrics_next` - A new module to support an alternate workflow to collect per-enrollment metrics and provide aggregate progress metrics. See commit, https://github.com/appsembler/figures/pull/429/commits/4dcbd75301545b2689f9079e5f6f8a30accfad31
* Update to `figures.pipeline.course_daily_metrics` to allow the deployment to chose which enrollment data collection and reporting to use. See commit, https://github.com/appsembler/figures/pull/429/commits/76aea1280500fd2da68dc4ba1a578b4b89611c2c
* Update to `figures.tasks` to provide new tasks to run this alternate workflow. See commit, https://github.com/appsembler/figures/pull/429/commits/2f087967d5cb5d2246a8bcd6637a1614d8e1aebf

See also https://github.com/appsembler/figures/issues/428

## What needs to follow on as a new PR

* Update the Django management commands to backfill data

### DIFF
--- a/figures/course.py
+++ b/figures/course.py
@@ -1,0 +1,115 @@
+"""Course specific module for Figures
+
+## This module defined a `Course` class for data retrieval
+
+Initialy created to do the following:
+
+1. Reduce duplication in Figures "pipeline"
+2. Build stronger course context to make Figures programming easier
+
+## Background summary
+
+A course id is globally unique as it has the identify of the organization and
+organizations are globally unique.
+
+## Design to think about - Enrollment class
+
+Build on Django's lazy eval for querysets to create also an `Enrollment` class
+that provides interfaces for `enrollment.date_for` and abstracts this bit of
+mess that enrollments and student modules do NOT associate and instead we need
+to query back and forth with `user_id` and `course_id`.
+"""
+from __future__ import absolute_import
+from django.db.models import Q
+from figures.compat import CourseEnrollment, StudentModule
+from figures.helpers import (
+    as_course_key,
+    as_date,
+)
+from figures.sites import (
+    get_site_for_course,
+)
+
+
+class Course(object):
+    """Representation of a Course.
+
+    The impetus for this class was dealing with querying for course enrollment
+    and student module records for a specific course and for dates and date
+    ranges for the course
+
+    ## Architecture goal
+
+    **Start simple and don't build the kitchen sink into here right away just
+    because this class exists**
+
+    ## Data under consideration to have this class handle
+
+    * enrollments created on a date, before, after, between. However this would
+    just be a convenience as the `.enrollments` property returns a queryset that
+    can be filtered on `.created`
+    """
+    def __init__(self, course_id):
+        """
+        Initial version, we pass in a course ID and cast to a course key as an
+        instance attribute. Later on, add `CourseLike` to abstract course identity
+        so we can stop worrying about "Is it a string repretentation of a course or
+        is it a CourseKey?"
+        """
+        self.course_key = as_course_key(course_id)
+
+        # Improvement: Consider doing lazy evaluation
+        self.site = get_site_for_course(self.course_id)
+
+    def __repr__(self):
+        return '{}.{} <{}>'.format(self.__module__,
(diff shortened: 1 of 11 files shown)


figures/course.py
```diff
@@ -0,0 +1,120 @@
...
+        return '{}.{} <{}>'.format(self.__module__,
+                                   self.__class__.__name__,
+                                   str(self.course_key))
+
+    def __str__(self):
+        return self.__repr__()
+
+    @property
+    def course_id(self):
+        """Returns string reprentation of the course id
```
johnbaldwin, line 79: Typo I need to fix > representation

figures/course.py
```diff
@@ -0,0 +1,120 @@
...
+
+    ## Architecture goal
+
+    **Start simple and don't build the kitchen sink into here right away just
+    because this class exists**
+
+    If we're serious about this class, do these:
+
+    * Lazy evaluation of self.site
+    * Add __str__, __repr__
```
johnbaldwin, line 49: Already added

figures/tasks.py
```diff
@@ -258,6 +263,64 @@ def populate_daily_metrics(site_id=None, date_for=None, force_update=False):
...
+        logger.info(msg.format(prefix=FPD_LOG_PREFIX,
+                               id=site.id,
+                               domain=site.domain,
+                               i=i,
+                               n=sites_count))
+        try:
+            populate_daily_metrics_for_site(site_id=site.id,
+                                            date_for=date_for,
+                                            ed_next=True,
+                                            force_update=force_update)
```
OmarIthawi, line 310: Celery queue question: This would run `populate_daily_metrics_for_site` synchronously. Do we want this behavior?

`populate_daily_metrics_for_site.delay()` should be easier to work with in terms of deployments and queue impact.

We have the `group()` and `delay()` pattern used in figures:

https://github.com/appsembler/figures/blob/2f087967d5cb5d2246a8bcd6637a1614d8e1aebf/figures/tasks.py#L456-L468

figures/pipeline/course_daily_metrics.py
```diff
@@ -235,16 +239,32 @@ class CourseDailyMetricsExtractor(object):
     BUT, we will then need to find a transform
     """
 
-    def extract(self, course_id, date_for, **_kwargs):
-        """
-            defaults = dict(
+    def extract(self, course_id, date_for, ed_next=False, **_kwargs):
```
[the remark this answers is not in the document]
johnbaldwin: Good point, I will do this!

figures/tasks.py
johnbaldwin: @OmarIthawi I _would_ prefer to run the tasks synchronously. We have not historically because when I originally implemented the celery tasks, I tried it synchronously with very simple vanilla tasks on standalone edx-platform and they would just "poof" disappear. No logs, no exceptions, nothing. Just drop out of existence. So we carried them on as serial executions since.

Soon in my queue is to drastically improve Figures backfill / data repair toolkit and health checking, and improve the Figures Celery docker devsite environment.

Once we have those plus documentation and training the platform team on how to use it. THEN we can optimize Figures tasks to run in parallel as you mention

figures/pipeline/course_daily_metrics.py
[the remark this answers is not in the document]
johnbaldwin: Hmm, looks like I already set `ed_next=False` as default. I agree, the name isn't great, but I did that abbreviation on purpose. Having you have a nit about it is a bonus. Once we accept "Workflow Next" as the new workflow for Tahoe, we can clean house on the old workflow, reduce a ton of code and this will go away too

PR #449, opened 2022-03-11
Enrollment data handling fixes

johnbaldwin:
This PR has fixes for the new enrollment data processing

### Fix backfill_enrollment_data_for_course task.Commit e3faba15cca87bc204b8eed55f3949b674c9ee25

* Problem was that `update_enrollment_data_for_course` returns a list of tuples. Each tuple is `(object, created)`, where `object` is the EnrollmentData record and `created` tells if the record was created
* Added logging to the backfill task
* Updated tests

### Fix class for Python27 Ginkgo. Commit  bf8340830c6bdf121597c8368ea0c5cd856f25b2

* one liner makes test class inherit from `object`, since we _still_ need to support py27.

### Adds enrollments filter method to Figures Course class. Commit bd0384242a7c2d6d7a102c66a10add3f16fdaa7a
* New method `figures.course.Course.enrollments_with_student_modules`
* This method finds `CourseEnrollment` records for a course for enrollments that have `StudentModule` records in the course. This functionality helps simplify finding out of date EnrollmentData records.

### Add test case to tests/test_course.py . Commit: 2566818c1937a408a617136e1de4aab9dc2fe5de

* Was missing the test case when there were no student modules in the course. We want to make sure our "empty set" handling is correct

### Add stale course enrollment handling to pipeline. Commit ded1349ac2ba85a131690b7c37e96c3b1cd796a1

* What's included here is a generator to yield CourseEnrollment records that need to have their EnrollmentData records updated
* If we install Figures on a site that's already been running or we  miss a day or more in processing, then the daily processing will not pick up on all "stale enrollments", enrollments that Figures EnrollmentData record updated.
* This is because Figures daily metrics jobs only look at "yesterday" in order to work around the limitations of StudentModule.
* The real logic is in `figures.enrollment.is_enrollment_data_out_of_date` (Which is still pretty simple)

### Fixed figures.tasks.backfill_enrollment_data_for_course. Commit 3511f2b2b44198b411655e1e9e8b063c57b5fbe9

* Now it looks for stale course enrollments and for each stale course enrollment found, updates that enrollment's metrics. Prior to this commit, it was just calling the normal daily enrolment data update function.
* Updated unit tests

### Update backfill enrollment management command. Commit cdeb6a299b085b983e73a65e4c34b8046a6c674c

* Add ability to load course ids from a file
* Add test coverage

One last thing to check: The student module GTE check in `figures.enrollment.student_modules_for_enrollment_after_date`

### DIFF
--- a/devsite/requirements/ginkgo.txt
+++ b/devsite/requirements/ginkgo.txt
@@ -83,3 +83,6 @@ freezegun==0.3.12
 
 # Added to address: TypeError: attrib() got an unexpected keyword argument 'convert'
 attrs==19.1.0
+
+# Added to enable mock.mock_open using 'builtins.open'
+future==0.18.2
--- a/devsite/requirements/hawthorn_base.txt
+++ b/devsite/requirements/hawthorn_base.txt
@@ -95,3 +95,6 @@ edx-lint==0.5.5
 
 # To address: edx-lint 0.5.5 requires astroid==1.5.2, but you'll have astroid 1.6.6 which is incompatible.
 astroid==1.5.2
+
+# Added to enable mock.mock_open using 'builtins.open'
+future==0.18.2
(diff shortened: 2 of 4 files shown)


OmarIthawi (approved):

johnbaldwin:
Thanks @OmarIthawi !

PR #458, opened 2022-04-26
Update SDM historic course count calculation

johnbaldwin:
* Old system wasn't working, was getting course counts as of the current
date
* Added code to estimate when a course was started from its first
enrollment created date
* So basically we now tracking counts of when a course had it first
enrollment and basing historical course count as of a date based on that
* This is to improve accuracy for Figures SDM backfill


OmarIthawi (approved):

johnbaldwin:
Thanks @OmarIthawi !

PR #460, opened 2022-05-31
Fix broken devsite data seeding

johnbaldwin:
* `devsite.seed` was not updated when `figures.backfill` was moved to `figures.pipeline.backfill`. This fixes that issue
* `devsite.settings` needed to have `FIGURES` key preset. This was causing devsite seed to fail too
* Improved the user creation integrity error by displaying the email address. Users are dynamically generated with `Faker` and sometimes the same email address is used. While not strictly needed, it may help the developer experience a little bit by seeing what email generated caused the collision


OmarIthawi (approved):

johnbaldwin:
Thanks @OmarIthawi !

PR #462, opened 2022-07-01
Add logging statement at start of the per-site SMM task

johnbaldwin:
Purpose is to be able to check the log files if the monthly subtask was started and when


bryanlandia (approved):

johnbaldwin:
Thanks, guys!

PR #463, opened 2022-08-12
Fix tasks for standalone and fix average progress validation error

johnbaldwin:
https://appsembler.atlassian.net/browse/BLACK-2554

## Fix average progress validation error

The problem was that the new pipeline workflow did not round enrollments' "progress percent" when averaging for CourseDailyMetrics average_progress fixed floating point field. This commit addresses this

## Fix site handling for standalone mode

This commit makes figures.sites.get_sites aware of standalone mode. The purpose of this is to help make standalone and multisite modes transparent to the bulk of Figures code
    
There are also a couple of minor changes in figures.tasks
    
1. populate_daily_metrics_next now includes the exception string to the error log entry
2. run_figures_monthly_metrics is now standalone aware. Since only one site is run for standalone, there's no need to spawn the site level function as a subtask

After this PR, I'll  do a PR bump for version to 0.4.3

### DIFF
--- a/figures/pipeline/enrollment_metrics_next.py
+++ b/figures/pipeline/enrollment_metrics_next.py
@@ -29,6 +29,7 @@ This module improves on the existing enrollment metrics collection module,
 This module provides
 
 """
+from decimal import Decimal
 from django.db.models import Avg
 from figures.course import Course
 from figures.enrollment import is_enrollment_data_out_of_date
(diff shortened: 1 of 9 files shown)


OmarIthawi (approved):

johnbaldwin:
@OmarIthawi Thanks for approving!

PR #470, opened 2022-09-26
run separate tasks per site in daily jobs

melvinsoft:
## Change description

This PR introduces a new version of the function `populate_daily_metrics`, this time what it does, instead of running the whole process inside a unique celery tasks, it creates a single tasks per each site and throw them in a celery chord. I know the change is no the ultimate solution, since site can have really different sizes, but still, having separate celery tasks per site, will allow us for example, to run production deployments at any time.

I'm also planning to to create a new queue for figures with a good concurrency (around 20) which on top on John's recent improvements (lower down the processing time from 10 to 2hs) should make the pipeline really fast to complete.

The PR also adds a new version of `populate_daily_metrics_for_site` called `populate_daily_metrics_for_site_async_sites`.

I already tested the code in Staging pulling it manually and it works, but I'd like to configure the new function and have it running for  a few days in Staging.

## Type of change
- [ ] Bug fix (fixes an issue)
- [x] New feature (adds functionality)

## Related issues

No relates issues

### Development

- [x] Lint rules pass locally
- [x] Application changes have been tested thoroughly
- [x] Automated tests covering modified code pass

### Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

### Code review 

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached as necessary
- [x] "Ready for review" label attached and reviewers assigned
- [ ] Changes have been reviewed by at least one other contributor
- [ ] Pull request linked to task tracker where applicable

PR #471, opened 2022-09-27
run separate tasks per site in daily jobs

OmarIthawi:
## Change description (re-opens #470)

This PR introduces a new version of the function `populate_daily_metrics`, this time what it does, instead of running the whole process inside a unique celery tasks, it creates a single tasks per each site and throw them in a celery chord. I know the change is no the ultimate solution, since site can have really different sizes, but still, having separate celery tasks per site, will allow us for example, to run production deployments at any time.

I'm also planning to to create a new queue for figures with a good concurrency (around 20) which on top on John's recent improvements (lower down the processing time from 10 to 2hs) should make the pipeline really fast to complete.

The PR also adds a new version of `populate_daily_metrics_for_site` called `populate_daily_metrics_for_site_async_sites`.

I already tested the code in Staging pulling it manually and it works, but I'd like to configure the new function and have it running for  a few days in Staging.

## Type of change
- [ ] Bug fix (fixes an issue)
- [x] New feature (adds functionality)

## Related issues

No relates issues

### Development

- [x] Lint rules pass locally
- [x] Application changes have been tested thoroughly
- [x] Automated tests covering modified code pass

### Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

### Code review 

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached as necessary
- [x] "Ready for review" label attached and reviewers assigned
- [ ] Changes have been reviewed by at least one other contributor
- [ ] Pull request linked to task tracker where applicable

### DIFF
--- a/figures/tasks.py
+++ b/figures/tasks.py
@@ -132,6 +132,52 @@ def populate_daily_metrics_for_site(site_id, date_for, ed_next=False, force_upda
 
 
 @shared_task
+def populate_daily_metrics_for_site_async_sites(site_id, date_for,
+                                                ed_next=False, force_update=False):
+    """Collect metrics for the given site and date,
+    """
+    try:
+        site = Site.objects.get(id=site_id)
+    except Site.DoesNotExist as e:
+        msg = ('{prefix}:SITE:FAIL:populate_daily_metrics_for_site:site_id: '
+               '{site_id} does not exist')
+        logger.exception(msg.format(prefix=FPD_LOG_PREFIX, site_id=site_id))
+        raise e
+
+    msg = '{prefix}:SITE:START:{id}:{domain}'
+    logger.info(msg.format(prefix=FPD_LOG_PREFIX,
+                           id=site.id,
+                           domain=site.domain))
+
+    for course_id in site_course_ids(site):
+        try:
+            if ed_next:
+                update_enrollment_data_for_course(course_id)
+
+            populate_single_cdm(course_id=course_id,
+                                date_for=date_for,
+                                ed_next=ed_next,
+                                force_update=force_update)
+        except Exception as e:  # pylint: disable=broad-except
+            msg = ('{prefix}:SITE:COURSE:FAIL:populate_daily_metrics_for_site.'
+                   ' site_id:{site_id}, date_for:{date_for}. course_id:{course_id}'
+                   ' exception:{exception}')
+            logger.exception(msg.format(prefix=FPD_LOG_PREFIX,
+                                        site_id=site_id,
+                                        date_for=date_for,
+                                        course_id=str(course_id),
+                                        exception=e))
+    populate_single_sdm(site_id=site.id,
+                        date_for=date_for,
+                        force_update=force_update)
+
+    msg = '{prefix}:SITE:END:{id}:{domain}'
+    logger.info(msg.format(prefix=FPD_LOG_PREFIX,
+                           id=site.id,
+                           domain=site.domain))
+
+
+@shared_task
 def update_enrollment_data_for_site(site_id, **_kwargs):
     """Original task to collect `EnrollmentData` records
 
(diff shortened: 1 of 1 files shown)


figures/tasks.py
```diff
@@ -328,6 +374,63 @@ def populate_daily_metrics_next(site_id=None, force_update=False):
...
+    * Figures collects the enrollment data first, then aggregates daily data.
+
+    TODO: Draft up public architecture docs and reference them here
+    """
+    if waffle.switch_is_active(WAFFLE_DISABLE_PIPELINE):
+        logger.warning('Figures pipeline is disabled due to %s being active.',
+                       WAFFLE_DISABLE_PIPELINE)
+        return
+
+    date_for = datetime.datetime.utcnow().date()
```
thraxil, line 405: I guess I'll leave my usual comment that this function would be easier to test and more generally useful if `date_for` is passed as a parameter with default `None` and only set to the current date when unset, so:

```
def populate_daily_metrics_async_sites(site_id=None, force_update=False, date_for=None):
...
if not date_for:
    date_for = datetime.datetime.utcnow().date()
```

figures/tasks.py
```diff
@@ -328,6 +374,63 @@ def populate_daily_metrics_next(site_id=None, force_update=False):
...
+                           site_count=sites_count))
+
+    logger.info('Starting figures.tasks.populate_daily_metrics...')
+    msg_metrics_end = '{prefix}:END:date_for={date_for}, site_count={site_count}'.format(
+        prefix=FPD_LOG_PREFIX, date_for=date_for, site_count=sites_count)
+
+    daily_tasks_list = [
+        populate_daily_metrics_for_site_async_sites.s(
+            site_id=site.id,
+            date_for=date_for.strftime('%m/%d/%Y'),
```
thraxil, line 424: I find it a bit odd that we're serializing `date_for` into a string to pass to other functions. It seems like it would be better to keep it a `datetime` object and just make the rest of the code work that way. Doing it this way just seems to be asking for bugs with timezones, etc.

johnbaldwin:
Re-posting my comment from #470, 

One thing we will need to add is test coverage for the new task functions. Tests for the current daily functions is in here: https://github.com/appsembler/figures/blob/main/tests/tasks/test_daily_tasks.py

This also include logging checking with caplog (just search the test module for it), which you will probably want to test [daily_metrics_callback](https://github.com/appsembler/figures/pull/470/files#diff-40202ed33c4550f30f989722cfab049b08ea7e34621019ea5b0e656fca247f89R378)

figures/tasks.py
```diff
@@ -131,6 +131,52 @@ def populate_daily_metrics_for_site(site_id, date_for, ed_next=False, force_upda
                         force_update=force_update)
 
 
+@shared_task
+def populate_daily_metrics_for_site_async_sites(site_id, date_for,
+                                                ed_next=False, force_update=False):
+    """Collect metrics for the given site and date,
```
johnbaldwin, line 137: @OmarIthawi Would you please provide a summary in the docstring? Why this function? Is it for Tahoe only, can it run in standalone? What do you want to jog your memory if you have to revisit this function several months from now?

Thanks!

figures/tasks.py
```diff
@@ -328,6 +374,63 @@ def populate_daily_metrics_next(site_id=None, force_update=False):
                            site_count=sites_count))
 
 
+@shared_task
+def daily_metrics_callback(results, *args):
```
johnbaldwin, line 378: @OmarIthawi looks like just a logger function. Does it or will it serve any other purpose? If it is only for logging, perhaps add a logging context to the function name?

figures/tasks.py
johnbaldwin: Reviewing in more depth, it appears that this function duplicates `populate_daily_metrics_for_site`, https://github.com/appsembler/figures/blob/main/figures/tasks.py#L100

 with the only change to add in logging, right? If this is the case, then let's just add the logging to `populate_daily_metrics_for_site` universally. Then we shouldn't need this new function

figures/tasks.py
```diff
@@ -328,6 +374,63 @@ def populate_daily_metrics_next(site_id=None, force_update=False):
...
+    logger.info(results)
+    logger.info(args)
+
+
+@shared_task
+def populate_daily_metrics_async_sites(site_id=None, force_update=False):
+    """Next iteration to collect daily metrics for all sites in a deployment
```
johnbaldwin, line 385: @OmarIthawi This docstring appears unchanged from `populate_daily_metrics_next`. Please provide context, lifting my comment from above:

Why this function? What problem does it solve? Is it for Tahoe only, can it run in standalone? What do you want to jog your memory if you have to revisit this function several months from now or hand off to another engineer? (Stuff that might not be crystal clear in the function source

To note, I should plan to remove the original `populate_daily_metrics` method and revamp the docstring for `populate_daily_metrics_next` to answer the same questions

figures/tasks.py
johnbaldwin: Just to note, this would duplicate the `SITE:START` and `SITE:END` logging output when called by `populate_daily_metrics_next` _until_ we update the logging output in `populate_daily_metrics_next`. But I think this is acceptable for now

johnbaldwin (changes requested):
@OmarIthawi Overall, I like us moving in the direction to scale out Figures pipeline execution.

My main points

1. New code needs test coverage
2. It looks like we do not need the new `populate_daily_metrics_for_site_async_sites` function as it appears all it does is add the "SITE:START" and "SITE:END" logging, which we can just add to the exiting `populate_daily_metrics_for_site` function, even if this duplicates logging when running `populate_daily_metrics_next` I'm happy to follow up and fix that as well as address Anders' `date_for` comment in `populate_daily_metrics_next

OmarIthawi:
Closing at the moment but the task is still open: https://appsembler.atlassian.net/browse/RED-3411?searchSessionId=dd94623f-4347-4bcb-827f-dbe6dee164ba&searchObjectId=23801&searchContainerId=10012&searchContentType=issue

